PR #7, opened 2020-06-02
DM-25044: Initial development of MTCS class in ts_observatory_control

tribeiro:
Some minor changes to `RemoteGroup` to support components that only send out telemetry and events and do not reply to commands. This is to support the MTMount component.
Add `BaseGroupMock` class. This class will make writing of mock classes with group of CSCs slightly easier, by taking care of a the basics.
Modify atcs and latiss mock classes to use the BaseGroupMock.
Initial implementation of MTCS with mock class and unit tests. Currently implemented the basics and a couple of slew commands.
Update version history.


python/lsst/ts/observatory/control/mock/base_group_mock.py
```diff
@@ -0,0 +1,254 @@
...
+    This is useful for unit testing.
+
+    Parameters
+    ----------
+    components : `list` [`str`]
+        A list of strings with the names of the SAL components that are part
+        of the group.
+    output_only : `list` [`str`]
+        A list of strings with the names of the SAL components that only send
+        telemetry and events, but do not reply to commands.
```
r-owen, line 45: Please document that `None` is acceptable, e.g.
```
output_only  : `list` [`str`] or `None`
```

python/lsst/ts/observatory/control/mock/base_group_mock.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def __init__(self, components, output_only=()):
+
+        self._components = components
+
+        self.components = [comp.lower() for comp in self._components]
+
+        self._component_names = []
+
+        self._output_only = output_only if output_only is not None else []
```
r-owen, line 57: Please check that the entries in `output_only` exist in `components` in order to catch typos. (I'd probably use the `set` class for that).

python/lsst/ts/observatory/control/mock/base_group_mock.py
```diff
@@ -0,0 +1,254 @@
...
+        _controllers = {}
+        self._output_only = []
+
+        for i, component in enumerate(self._components):
+            name, index = salobj.name_to_name_index(component)
+            rname = component.lower().replace(":", "_")
+            self._component_names.append(rname)
+            _controllers[rname] = salobj.Controller(name=name, index=index)
+            if component in output_only:
+                self._output_only.append(rname)
```
r-owen, line 68: This looks like a mistake because you already set `self._output_only` above. Now you will have 2 copies of each entry.  I think you can just drop lines 67 and 68.

Also consider making `self._output_only` a `set` . It's more efficient to search and eliminates duplicates.

python/lsst/ts/observatory/control/mock/base_group_mock.py
```diff
@@ -0,0 +1,254 @@
...
+                ).cmd_disable.callback = self.get_disable_callback(comp)
+                getattr(
+                    self.controllers, comp
+                ).cmd_standby.callback = self.get_standby_callback(comp)
+
+        for comp in self.output_only:
+            for cmd in getattr(self.controllers, comp).salinfo.command_names:
+                getattr(
+                    getattr(self.controllers, comp), f"cmd_{cmd}"
+                ).callback = self.generic_raise_callback
```
r-owen, line 98: Nice

python/lsst/ts/observatory/control/mock/base_group_mock.py
```diff
@@ -0,0 +1,254 @@
...
+
+    async def generic_callback(self, data):
+        await asyncio.sleep(HEARTBEAT_INTERVAL)
+
+    async def generic_raise_callback(self, data):
+        """ A generic callback function that will raise an exception if called.
+        """
+        raise RuntimeError("This command should not be called.")
+
+    def get_start_callback(self, comp):
```
r-owen, line 138: Please consider rewriting these as the actual callback methods and using `functools.partial` to send the `comp` argument when you specify the method as the callback function.

I _think_ it would make the code significantly easier to read.

For example:
```
# Define the method
async def start_callback(self, data, comp):
    ss = getattr(self.controllers, comp).evt_summaryState
    ...
...
# Use the method:
getattr(
    self.controllers, comp
).cmd_start.callback = functools.partial(self.get_start_callback, comp=comp))
```

python/lsst/ts/observatory/control/mock/mtcs_mock.py
```diff
@@ -0,0 +1,326 @@
...
+from lsst.ts.idl.enums import MTPtg
+
+from ..utils import subtract_angles
+
+LONG_TIMEOUT = 30  # seconds
+HEARTBEAT_INTERVAL = 1  # seconds
+CLOSE_SLEEP = 5  # seconds
+
+
+class MTCSMock(BaseGroupMock):
```
r-owen, line 44: Would it make any sense to have a base class shared by ATCSMock and MTCSMock (that inherited fromBaseGroupMock)? I realize the components differ so there'd not be a huge overlap, but perhaps there is enough to be useful?

tests/maintel/test_mtcs.py
```diff
@@ -0,0 +1,178 @@
...
+            self.assertEqual(rot_rec.Demand, rot_set)
+
+    async def test_slew_all(self):
+
+        async with self.make_group(
+            usage=MTCSUsages.StartUp + MTCSUsages.Shutdown + MTCSUsages.Slew
+        ):
+
+            # enable all components
+
```
r-owen, line 145: Please consider eliminating the blank lines after the comments.

tests/maintel/test_mtcs.py
```diff
@@ -0,0 +1,178 @@
...
+            # This is a circumpolar object, should be always vizible in
+            # Pachon.
+
+            name = "HD 185975"
+            ra = Angle("20:28:18.74", unit=u.hourangle)
+            dec = Angle("-87:28:19.9", unit=u.deg)
+
+            # Test that slew command works with RA/Dec alone
+            with self.subTest("test_slew", ra=ra, dec=dec):
```
r-owen, line 156: I strongly suggest removing `subTest` as used here. It is intended for loops (repeated tests over different values) where it is not clear what the value was. In this case you only have the one case, so a failure will be obvious from the traceback.

PR #9, opened 2020-06-18
Add "persistent" option to `offset_azel` and `offset_xy` to support o…

tribeiro:
…ffsets that are retained between slews. Option still needs to be implemented.


python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -483,10 +488,22 @@ async def offset_azel(self, az, el, relative=False):
         relative : `bool`
             If `True` offset is applied relative to the current position, if
             `False` (default) offset replaces any existing offsets.
+        persistent : `bool`
+            Should the offset be absorbed and persist between slews?
```
wvreeven, line 492: "persist" -> "persisted"

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -503,18 +520,28 @@ async def offset_azel(self, az, el, relative=False):
...
         y : `float`
-        Offset in camera y-axis (arcsec).
+            Offset in camera y-axis (arcsec).
         relative : `bool`
-        If `True` offset is applied relative to the current position, if
-        `False` (default) offset replaces any existing offsets.
+            If `True` offset is applied relative to the current position, if
+            `False` (default) offset replaces any existing offsets.
+        persistent : `bool`
+            Should the offset be absorbed and persist between slews?
```
wvreeven, line 538: "persist" -> "persisted"

PR #11, opened 2020-08-03
Implement new documentation schema.

tribeiro:
Add user-manual section.
Add developer-guide section.
Add new feature to `RemoteGroup.set_state` to allow users to set the state of individual or a subgroup of CSCs.


doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+   from lsst.ts.observatory.control.auxtel.atcs import ATCS
+
+
+   class MyScript(salobj.BaseScript):
+
+     def __init__(self, index):
+
+         super().__init__(index=index, descr="My script for some operation.")
+
+         self.atcs = ATCS(domain=self.domain, log=self.log)
```
patrickingraham, line 82: You should probably show the instantiation of the log object?

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+
+.. _user-guide-limiting-resources:
+
+Limiting Resources
```
patrickingraham, line 86: should this go with the shared resources bit above?

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+The idea is that, after running :py:meth:`enable <lsst.ts.observatory.control.RemoteGroup.enable>`, all CSCs in the group will be in the ``ENABLED`` state.
+Components that where already in ``ENABLED`` state will be left in that state.
+For components in any other state the method will make sure to send the required state transition commands to enable them.
+The method can run without any input argument, though one must be aware that it need to know what setting to load when transition from ``STANDBY`` to ``DISABLED``.
+If no input is given, the method will inspect the ``settingVersions`` event from the CSCs in the group and select the first item in ``settingVersions.recommendedSettingsLabels``.
+It is possible to provide settings either partially of fully using the ``settings`` input parameter.
+
+Following are a couple of examples of how to use the :py:meth:`enable <lsst.ts.observatory.control.RemoteGroup.enable>` method.
+
+This will inspect the ``settingVersions`` event from all CSCs in the `ATCS` group to determine the ``settingsToApply`` for each one of them.
```
patrickingraham, line 144: I wish we could link to more information on configuration here.... but until we get the config stuff implemented i suppose it's not feasible. "How do I inspect the settingVersions" is the immediate question. Maybe put an example of how to print that info?

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+            "atmcs": "",
+            "atptg": "",
+            "atpneumatics": "",
+            "athexapod": "current",
+            "atdome": "test.yaml",
+            "atdometrajectory": "",
+        }
+    )
+
+Then, to send ``all`` components to ``STANDBY`` state;
```
patrickingraham, line 180: Remove the "Then" as it sounds like an instruction to perform after enabling :)

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+
+Furthermore, the TCS provides a suit of useful slew operations.
+
+To slew the telescope to an AzEl coordinate (for instance, to conduct some maintenance of calibration), it is possible to use :py:meth:`point_azel <lsst.ts.observatory.control.BaseTCS.point_azel>`.
+The method will slew to a fixed position in the local coordinate and `will not` initiate tracking.
+For instance, to slew the telescope to `azimuth=0.` and `elevation=80.` degrees;
+
+.. code:: python
+
+    await tcs.point_azel(az = 0., el=80.)
```
patrickingraham, line 265: so I think this might be a bit confusing. Maybe you want to show the instantiation of the tcs class showing which tcs you're using? Otherwise it sounds a bit like you're using the baseclass to do this.

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+    :target: https://github.com/lsst-ts/ts_observatory_control
+.. image:: https://img.shields.io/badge/Jenkins-ts_observatory_control-green.svg
+    :target: https://tssw-ci.lsst.org/job/LSST_Telescope-and-Site/job/ts_observatory_control/
+.. image:: https://img.shields.io/badge/Jira-ts_observatory_control-green.svg
+    :target: https://jira.lsstcorp.org/issues/?jql=labels+%3D+ts_observatory_control
+
+The Observatory Control package provides users and developers with tools to interact with the Rubin Observatory System.
+These tools are mainly Python classes/objects that represent major groups of components and encapsulate high-level operations involving those components.
+For instance, the `ATCS` class combine functionality for all "auxiliary telescope" components of the Vera Rubin Auxiliary Telescope.
+Actions like slewing the telescope, while making sure all involved components are in `ENABLED` state and waiting until all components are in position is encapsulate in a single command;
```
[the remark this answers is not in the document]
patrickingraham: ok, so after reading below I see there are tonnes of examples down there...

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+
+.. code:: python
+
+    await tcs.slew_icrs(
+              ra=20.0, dec=-80.0, rot_par=0., target_name="Test target"
+          )
+
+Although ``rot_par=0.`` is the most commonly used value, the user is free to select any angle.
+
+In case the user selects an angle that would cause the rotator to slew to a position outside the valid range, the task will fail and raise an exception.
```
patrickingraham, line 372: "In case the user demands an angle outside the valid range, the task will fail and raise an exception and not slew to the demanded position"

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+  <ipython-input-25-be270f3a125b> in async-def-wrapper()
+  .
+  .
+  .
+  AckError: msg='Command failed', ackcmd=(ackcmd private_seqNum=1597989109,
+  ack=<SalRetCode.CMD_FAILED: -302>, error=6611,
+  result='Rejected : rotator position angle out of range')
+
+The error message will show the exception traceback, which can be somewhat intimidating.
+However, the important bit of information can be found in the last couple lines of the output (as shown above).
```
patrickingraham, line 386: Is there something that indicates which CSC the error comes from? It'd be nice to show that as well since the exceptions aren't often thrown from the TCS class..

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+            "atmcs": "",
+            "atptg": "",
+            "atpneumatics": "",
+            "athexapod": "current",
+            "atdome": "test.yaml",
+            "atdometrajectory": "",
+        }
+    )
+
+It is important to remember that, if the components are already enabled, they will be left in the ``ENABLED`` state and will not be re-cycled.
```
patrickingraham, line 535: Uh oh, so what happens if something is enabled already, then you send the enabled command with a different configuration than what it is enabled with? It throws an exception I hope?

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+    from lsst.ts.observatory.control.auxtel import ATCS, ATCSUsages
+
+    atcs = ATCS(intended_usage=ATCSUsages.Slew)
+
+    await atcs.start_task
+
+    # Minimum set of parameters.
+    await atcs.slew_icrs(ra="00 42 44.330", dec="+41 16 07.50")
+
+    # Explicitly specify rot_sky and targer_name (both optional).
```
patrickingraham, line 552: targer = target

doc/user-guide/user-guide.rst
```diff
@@ -0,0 +1,598 @@
...
+
+    # Minimum set of parameters.
+    await atcs.slew_object("M31")
+
+    # Explicitly specify rot_sky (optional).
+    await atcs.slew_object("M31", rot_sky=0.)
+
+For shutting down the observatory :py:meth:`shutdown <lsst.ts.observatory.control.BaseTCS.shutdown>`
+
+.. _user-guide-latiss:
```
patrickingraham, line 565: so these are good but this page is going to get massive! 
we might want to think about how to divide it up

doc/user-guide/user-guide.rst
tribeiro: No, it is part of the Script class already... I'll mention it in the text...

doc/user-guide/user-guide.rst
tribeiro: This seems like a nice feature for the `RemoteGroup` let me add this as a new feature and document it later. I'll add a comment here to document once the feature exists.

doc/user-guide/user-guide.rst
tribeiro: I added a note before about that... hope it helps ease the confusion.

doc/user-guide/user-guide.rst
tribeiro: uhm... that is a good point... unfortunately, I don't think that information is readily available... there are a couple ways we could go about it either salobj can be more explicit about who rejects the command or the module could be... not sure what would be best.... I think salobj philosophy is that the user knows to which CSC it sent the command to, but this is not true in this case...

doc/user-guide/user-guide.rst
tribeiro: no... it is just ignored because the CSC is already enabled...

doc/user-guide/user-guide.rst
tribeiro: yeah...

PR #12, opened 2020-08-21
Reformatted (split up) documentation, specifically the user guide. Gr…

patrickingraham:
…ouped classes according to telescope, also make one page per class.

### DIFF
--- a/doc/user-guide/maintel/mtcs-user-guide.rst
+++ b/doc/user-guide/maintel/mtcs-user-guide.rst
@@ -1,0 +1,29 @@
+
+.. _user-guide-mtcs:
+
+Main Telescope Control System (MTCS)
+------------------------------------
+
+.. _Dome: https://ts-dome.lsst.io/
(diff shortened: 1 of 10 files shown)


doc/user-guide/maintel/mtcs-user-guide.rst
```diff
@@ -0,0 +1,29 @@
...
+The :py:class:`MTCS <lsst.ts.observatory.control.maintel.MTCS>` class groups the components that are related to Main Telescope operations, such as slewing and tracking objects.
+It is built on top of the :py:class:`BaseTCS <lsst.ts.observatory.control.BaseTCS>`.
+
+The component CSCs that are part of this group are:
+
+  * MTPtg
+  * `MTMount`_
+  * `MTAOS`_
+  * `MTDomeTrajectory`_
+  * `MTDome`_
```
tribeiro, line 24: I know we are renaming Dome to be MTDome but I think it would be better to wait until the renaming happens before adopting it do the docs.

PR #16, opened 2020-09-25
Tickets/dm 26495

patrickingraham:


python/lsst/ts/observatory/control/auxtel/latiss.py
```diff
@@ -266,6 +266,29 @@ async def setup_atspec(self, filter=None, grating=None, linear_stage=None):
             async with self.cmd_lock:
                 return await asyncio.gather(*setup_coroutines)
 
+    async def get_setup(self):
+        """Get the current filter, grating and stage position
+
+        Returns
+        -------
+        filter name : ``evt_reportedFilterPosition.name``
```
tribeiro, line 274: These names should be like variable names, e.g.`filter_name` also, the return type should be the actual return type so `string` in this case and I think `float` for the linear stage.

python/lsst/ts/observatory/control/auxtel/latiss.py
```diff
@@ -301,7 +324,9 @@ def usages(self):
                 components_attr=["atspectrograph"],
                 readonly=False,
                 generics=["summaryState"],
-                atspectrograph=["changeFilter", "changeDisperser", "moveLinearStage"],
+                atspectrograph=["changeFilter", "changeDisperser", "moveLinearStage",
```
tribeiro, line 327: Can you add this to the `self.valid_use_cases.All` and `self.valid_use_cases.TakeImageFull` usages?

python/lsst/ts/observatory/control/constants/atcs_constants.py
```diff
@@ -0,0 +1 @@
+plate_scale = 0.01
```
tribeiro, line 1: Can you a comment about the units? or, consider using satrapy units to define the variable.

python/lsst/ts/observatory/control/constants/latiss_constants.py
```diff
@@ -0,0 +1,10 @@
+"""Useful constants for running auxTel"""
+from lsst.geom import ExtentD, PointD
```
tribeiro, line 2: I think you are not using `ExtendD`

python/lsst/ts/observatory/control/constants/latiss_constants.py
```diff
@@ -0,0 +1,10 @@
...
+
+plate_scale = 0.1045  # exp.getWcs().getPixelScale().asArcseconds()  # arcsec/pixel
```
tribeiro, line 4: Please, remove `# exp.getWcs().getPixelScale().asArcseconds() ` from the comment

tests/auxtel/test_latiss.py
```diff
@@ -71,7 +71,9 @@ async def test_take_flats(self):
             nflats = 4
             exptime = 1.0
             filter_id = 1
+            filter_name = f'filter{filter_id}'
```
tribeiro, line 74: Do you mind using double-quotes? `f"filter{filter_id}"`, it is what is used throughout the code. 😬

python/lsst/ts/observatory/control/mock/latiss_mock.py
```diff
@@ -115,20 +117,23 @@ async def cmd_changeFilter_callback(self, data):
         await asyncio.sleep(0.1)
         self.atspec.evt_filterInPosition.put()
         self.atspec.evt_reportedFilterPosition.put()
+        self.atspec.evt_reportedFilterPosition.set_put(position=data.filter, name=f'filter{data.filter}')
```
tribeiro, line 120: Do you mind using double-quotes?

python/lsst/ts/observatory/control/mock/latiss_mock.py
```diff
@@ -115,20 +117,23 @@ async def cmd_changeFilter_callback(self, data):
...
         self.latiss_filter = data.filter
 
     async def cmd_changeDisperser_callback(self, data):
         """Emulate change filter command"""
         await asyncio.sleep(0.1)
         self.atspec.evt_disperserInPosition.put()
         self.atspec.evt_reportedDisperserPosition.put()
+        self.atspec.evt_reportedDisperserPosition.set_put(position=data.disperser,
+                                                          name=f'grating{data.disperser}')
```
tribeiro, line 129: Do you mind using double quotes?

tribeiro (changes requested):
Made a couple comments in line. Some are picky, I know... feel free to ignore the picky ones...

python/lsst/ts/observatory/control/auxtel/latiss.py
patrickingraham: note that the names aren't the *actual* names of the variables that are returned... which I think it was you mean?

tribeiro: yes, I noticed... but I mean, they should be valid variable names. If you don't want to document variable names you can document the type only, e.g.:

```
Returns
-------
int
    Description of anonymous integer return value.
```

https://numpydoc.readthedocs.io/en/latest/format.html

python/lsst/ts/observatory/control/mock/latiss_mock.py
patrickingraham: cruel..... but ok

python/lsst/ts/observatory/control/auxtel/latiss.py
patrickingraham: code returned pieces of objects, so rather than putting those pieces in variables and then returning them i just followed the example above.

tribeiro (approved):
Thanks for implementing the requested changes.

PR #22, opened 2021-01-14
DM-28187: Update ts_observatory_control and ts_standardscript to remove NewMTMount

tribeiro:
Implement changes required by xml 7.1:
	* Removes NewMTMount (replaced by MTMount)
	* Update MTMount topics names and attributes.
Improve error messages when heartbeat monitor fails.
Improve error messages when slew/track target commands fails.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -359,9 +359,13 @@ class ATCS(BaseTCS):
         else:
             # TODO: DM-24525 remove hard coded 8 when real ATMCS is fixed.
             while at_mount_state.state not in (8, ATMCS.AtMountState.TRACKINGDISABLED):
-                self.log.debug(
-                    f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}."
+                tracking_state = (
+                    ATMCS.AtMountState(at_mount_state.state)
+                    if at_mount_state.state in [value for value in ATMCS.AtMountState]
+                    else f"Unknown({at_mount_state.state})"
                 )
+
+                self.log.debug(f"Tracking state: {tracking_state!r}.")
                 at_mount_state = await self.rem.atmcs.evt_atMountState.next(
                     flush=False, timeout=self.long_timeout
                 )
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -359,13 +359,12 @@ class ATCS(BaseTCS):
         else:
             # TODO: DM-24525 remove hard coded 8 when real ATMCS is fixed.
             while at_mount_state.state not in (8, ATMCS.AtMountState.TRACKINGDISABLED):
-                tracking_state = (
-                    ATMCS.AtMountState(at_mount_state.state)
-                    if at_mount_state.state in [value for value in ATMCS.AtMountState]
-                    else f"Unknown({at_mount_state.state})"
-                )
-
-                self.log.debug(f"Tracking state: {tracking_state!r}.")
+                try:
+                    self.log.debug(
+                        f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}"
+                    )
+                except ValueError:
+                    self.log.warning(f"Unknown tracking state: {at_mount_state.state}")
                 at_mount_state = await self.rem.atmcs.evt_atMountState.next(
                     flush=False, timeout=self.long_timeout
                 )
--- a/python/lsst/ts/observatory/control/base_tcs.py
+++ b/python/lsst/ts/observatory/control/base_tcs.py
@@ -125,6 +125,11 @@ class BaseTCS(RemoteGroup, metaclass=abc.ABCMeta):
                 getattr(self.rem, self.ptg_name).cmd_azElTarget,
                 slew_timeout=slew_timeout,
             )
+        except salobj.AckError as ack_err:
+            self.log.error(
+                f"Command to slew to azEl target rejected: {ack_err.ackcmd.result}"
+            )
+            raise ack_err
         finally:
             self.unset_azel_slew_checks(check)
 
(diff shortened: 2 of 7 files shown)


python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -359,9 +359,13 @@ async def stop_tracking(self):
         else:
             # TODO: DM-24525 remove hard coded 8 when real ATMCS is fixed.
             while at_mount_state.state not in (8, ATMCS.AtMountState.TRACKINGDISABLED):
-                self.log.debug(
-                    f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}."
+                tracking_state = (
+                    ATMCS.AtMountState(at_mount_state.state)
+                    if at_mount_state.state in [value for value in ATMCS.AtMountState]
+                    else f"Unknown({at_mount_state.state})"
```
r-owen, line 365: This sets tracking_state to a string instead of an enum if the state is unrecognized, which makes the code harder to understand and maintain.

Since you don't actually use the state for anything, consider something like this:
```
try:
    self.log.debug(f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}") 
except ValueError:
    self.log.warning(f"Unknown tracking state {at_mount_state.state}")
```
with obvious modifications if you want to use the value for something.

If you are keen to test against a collection of valid values then I suggest creating it once in the constructor, as a frozenset: `self.valid_at_mount_states = frozenset(ATMCS.AtMountState)`. There is no need for explicit iteration.

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -125,6 +125,11 @@ async def point_azel(
                 getattr(self.rem, self.ptg_name).cmd_azElTarget,
                 slew_timeout=slew_timeout,
             )
+        except salobj.AckError as ack_err:
+            self.log.error(
+                f"Command to slew to azEl target rejected: {ack_err.ackcmd.result}"
+            )
+            raise ack_err
```
r-owen, line 132: This makes me a bit uneasy. It adds a lot of boilerplate if you will eventually want this for every command in this package. Usually the caller handles exceptions. Apparently that's not good enough here, but why? Perhaps RemoteCommand should log failed commands?

python/lsst/ts/observatory/control/mock/mtcs_mock.py
```diff
@@ -151,57 +149,51 @@ async def mount_telemetry(self):
...
                 if self.acting:
-                    self.controllers.newmtmount.evt_axesInPosition.set_put(
+                    self.controllers.mtmount.evt_axesInPosition.set_put(
                         elevation=in_position_elevation, azimuth=in_position_azimuth
                     )
 
-                self.controllers.mtmount.tel_Azimuth.set_put(
-                    Azimuth_Angle_Actual=az_actual + az_induced_error + az_dif.deg / 1.1
+                self.controllers.mtmount.tel_azimuth.set_put(
+                    angleActual=az_actual + az_induced_error + az_dif.deg / 1.1
```
r-owen, line 190: Pre-existing issue but please add a brief code comment about the constant 1.1 (and consider a named constant for clarity).

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -475,16 +475,21 @@ async def check_comp_heartbeat(self, component):
...
-                flush=True, timeout=self.fast_timeout
-            )
+            try:
+                await getattr(self.rem, component).evt_heartbeat.next(
+                    flush=True, timeout=self.fast_timeout
+                )
+            except asyncio.TimeoutError:
+                raise RuntimeError(
+                    f"No heartbeat from {component} received in {self.fast_timeout}s."
+                )
```
r-owen, line 492: Why this change? It seems unrelated to the ticket and seems unnecessary -- TimeoutError is the natural exception to raise. If you merely want to add a message then you can raise a new instance of asyncio.TimeoutError.

python/lsst/ts/observatory/control/utils/remote_group_test_case.py
```diff
@@ -146,6 +146,8 @@ async def wait_for(self, coro, timeout, description, verbose):
             print(f"wait for {description}")
         try:
             result = await asyncio.wait_for(coro, timeout=timeout)
+        except asyncio.CancelledError:
+            print(f"{description}: Cancelled error.")
```
r-owen, line 150: I suggest re-raising the cancelled error, as that more closely mimics `asyncio.wait_for`. Or if that causes problems then please explain in the doc string what you do with it.

r-owen (approved):
I have a few suggested changes and clarifications. Nothing major. Please look them over and decide what to do.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: great suggestion! thanks!

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: So this is a complain I got from some users. The exception is too much to parse for people that are not used to tracebacks. I don't know about logging it from the RemoteCommand. Might be a good idea if the `ackcmd.result` contains enough information.

python/lsst/ts/observatory/control/mock/mtcs_mock.py
tribeiro: Ok, I added a constant and comment with clarification.

python/lsst/ts/observatory/control/remote_group.py
tribeiro: The original exception doesn't say which component missed the heartbeat, so it is hard to identify the source of the problem. I will re-raise `asyncio.TimeoutError`.

python/lsst/ts/observatory/control/utils/remote_group_test_case.py
tribeiro: I think this got here by mistake... I was debugging a `CancelledError` exception that I get from time to time and I can't figure out where it is coming from.

r-owen: Ah, OK. I always try to remember to catch CancelledError if I'm catching other errors as well, since I usually want to treat cancellation specially (e.g. quit a loop, which might continue for other errors). So I assumed you were doing that here.

python/lsst/ts/observatory/control/remote_group.py
r-owen: Sounds good. Thank you for the explanation.

python/lsst/ts/observatory/control/base_tcs.py
r-owen: Oy. OK. I'll think about logging in RemoteCommand.

PR #23, opened 2021-01-25
DM-28417: Support AT observing run 2020/01

tribeiro:
Fix error with atcs.stop_tracking.
Make some improvement to ATCS open/close dome methods.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -359,9 +359,12 @@ class ATCS(BaseTCS):
         else:
             # TODO: DM-24525 remove hard coded 8 when real ATMCS is fixed.
             while at_mount_state.state not in (8, ATMCS.AtMountState.TRACKINGDISABLED):
-                self.log.debug(
-                    f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}."
-                )
+                try:
+                    self.log.debug(
+                        f"Tracking state: {ATMCS.AtMountState(at_mount_state.state)!r}."
+                    )
+                except ValueError:
+                    self.log.debug(f"Unknown tracking state: {at_mount_state.state}.")
                 at_mount_state = await self.rem.atmcs.evt_atMountState.next(
                     flush=False, timeout=self.long_timeout
                 )
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -556,25 +556,52 @@ class ATCS(BaseTCS):
 
     async def open_dome_shutter(self):
         """Task to open dome shutter and return when it is done.
+
+        Raises
+        ------
+        RuntimeError:
+            If mainDoorState.state is neither OPENED nor CLOSED
+
         """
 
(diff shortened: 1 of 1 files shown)


python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -553,25 +556,52 @@ async def shutdown(self):
...
+                (
+                    await self.rem.atdome.evt_mainDoorState.aget(
+                        timeout=self.fast_timeout
+                    )
+                ).state
+            )
+
+            if shutter_pos != ATDome.ShutterDoorState.OPENED:
+                self.log.debug(
+                    f"Open shutter command returned while shutter still {shutter_pos!r}."
```
patrickingraham, line 589: why does this happen? is it a bug that needs fixing in ATDome? should we be referencing a ticket number?

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -594,26 +624,56 @@ async def home_dome(self):
...
-            ]
+            shutter_pos = ATDome.ShutterDoorState(
+                (
+                    await self.rem.atdome.evt_mainDoorState.next(
+                        flush=False, timeout=self.open_dome_shutter_time
+                    )
+                ).state
+            )
+
+            if shutter_pos != ATDome.ShutterDoorState.CLOSED:
```
patrickingraham, line 658: same comment

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -908,9 +1007,16 @@ def flush_offset_events(self):
...
+        while True:
+            in_position = await self.rem.atmcs.evt_allAxesInPosition.next(
+                flush=False, timeout=self.tel_settle_time
+            )
+
+            if in_position.inPosition:
+                self.log.debug("All axes in position.")
+                return in_position.inPosition
+            else:
+                self.log.debug("Telescope not in position.")
```
patrickingraham, line 1019: should this not raise an error?

patrickingraham (changes requested):
code itself is fine, but i'm missing the context as to why it needs to exist...
Maybe just needs a comment or two explaining why?

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: No, if `in_position.inPosition = False` it means the telescope is not in position *yet*, so then you proceed to wait for the event again, until it is `True`. Also, note that, when we send an offset command the mount will first issue an `in_position.inPosition=False` (when the offset start) then `in_position.inPosition=True` (when the offset is complete).

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: This (and the next one) is to handle the bug in ATDome. Yes, this should be removed once ATDome is fixed, *but* at the same time, this is not the main code base. It is a patch for "release/0.4", to work with the observatory during the run.

patrickingraham: ah i missed it was the patch. Sorry

patrickingraham (approved):
Comments addressed. Approved

PR #24, opened 2021-02-05
tickets/DM-28673

mareuter:


python/lsst/ts/observatory/control/auxtel/latiss.py
```diff
@@ -157,6 +157,8 @@ async def expose(
             key_value_map = f"groupId: {group_id},imageType: {image_type}"
             if test_type is not None:
                 key_value_map += f",testType: {test_type}"
+            else:
+                key_value_map += f",testType: {image_type}"
```
patrickingraham, line 161: Consider adding a comment why you need to do this. Also comment regarding update of docstring

patrickingraham (approved):
Code is fine, but consider adding in the docstring that the field gets auto populated?
Might be worth adding a comment (with the new code) adding why it is necessary to populate that field.
I'm actually unsure myself as I was under the impression it was optional?

python/lsst/ts/observatory/control/auxtel/latiss.py
mareuter: I'll add some comments. It's not optional thanks to the gen3 ingest. Learned this the hard way on ComCam.

tribeiro:
@mareuter can we get this PR rebased and merged? it is getting stale here.

mareuter:
Withdrawing change.

PR #25, opened 2021-02-09
DM-28456: Bug fixes in the upstream versions of ts_observatory_control and ts_standardscripts

tribeiro:
* Implement workaround for issue with ATDome not reliably finishing open/close dome commands.
* Fix offset_done method in ATCS, to properly wait for offset to be completed.
* Add boresight xy-axis parity determination in ATCS.
* Improve handling of check.<component> in ATCS.shutdown.
* Implement xml 8 backward compatibility for MTMount in MTCS.
* Add scripts to run mocks from the command line.


bin/run_atcs_mock.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+# This file is part of ts_observatory_control.
+#
+# Developed for the LSST Telescope and Site.
+# This product includes software developed by the LSST Project
```
wvreeven, line 5: Please refer to `Vera Rubin Observatory` instead of `LSST` in this and all other files.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -498,83 +499,131 @@ async def shutdown(self):
...
 
-        try:
-
-            await self.rem.ataos.cmd_disableCorrection.set_start(
-                disableAll=True, timeout=self.long_timeout
-            )
-        except Exception:
-            self.log.exception("Failed to disable ATAOS corrections. Continuing...")
+        if check.ataos:
+            self.log.debug(f"Disabling ATAOS corrections.")
```
wvreeven, line 503: I find it confusing that a boolean called `check.ataos` which evaluates to `true` leads to disabling corrections. Shouldn't this boolean be `false` instead?

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -628,6 +727,66 @@ async def close_dome(self):
                 f"{ATDome.ShutterDoorState.CLOSED}"
             )
 
+    async def _wait_for_shutter_door_state(self, state, cmd_task=None, timeout=None):
+        """Waits for `ATDome.ShutterDoorState.state` to match a required
```
wvreeven, line 731: Please use an active voice for doc strings, i.e. `Wait for` instead of `Waits for`.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -889,31 +1048,90 @@ async def _slew_to(
...
+            else -np.mean(nasmyth.nasmyth2CalculatedAngle)
         )
 
+        angle = np.mean(azel.elevationCalculatedAngle) + nasmyth_angle + 90.0
+
         return angle
 
     def flush_offset_events(self):
-        """Abstract method to flush events before and offset is performed.
+        """Implement abstract method to flush events before and offset is
```
wvreeven, line 1089: Should `and` be `an` here?

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -889,31 +1048,90 @@ async def _slew_to(
...
+        offset_azel : Offset in local AzEl coordinates.
+        offset_xy : Offset in terms of boresight.
+        offset_radec : Offset in sky coordinates.
+
         """
         self.rem.atmcs.evt_allAxesInPosition.flush()
 
     async def offset_done(self):
-        """Wait for offset events.
+        """Wait for events specifying that and offset completed.
```
wvreeven, line 1103: Should `and` be `an` here?

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1279,6 +1503,7 @@ def usages(self):
                     "stopTracking",
                     "offsetAzEl",
                     "offsetRADec",
+                    "poriginOffset",
```
wvreeven, line 1506: Was this intentionally spelled like this or should this be `originOffset`? If intentionally then please consider naming it `pOriginOffset` instead here and everywhere else.

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -76,6 +76,14 @@ def __init__(self, components, domain=None, log=None, intended_usage=None):
         # it should read this from events.
         self.rotator_limits = [-90.0, +90.0]
 
+        # Parity of x and y axis. These can be 1 or -1 depending on how the
+        # x axis in the boresight is aligned with the telescope axis. For
+        # instance, Nasmyth angle right as parity 1 and Nasmyth angle left has
```
wvreeven, line 81: `as` should be `has`.

wvreeven (changes requested):
Several comments for your consideration. Also the Jenkins build failed.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: That is because this is a "shutdown" method... the `check` attribute is a way to skip certain components from general operations in case they are not available or something like that. I should probably document this somewhere.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: uhm.. interesting point... unfortunately, this is the name of the command in the xml and I can't change it.

python/lsst/ts/observatory/control/auxtel/atcs.py
wvreeven: Ah, sorry, I missed that. Thanks for the explanation.

python/lsst/ts/observatory/control/auxtel/atcs.py
wvreeven: Maybe something for a future ticket?

wvreeven (approved):
Thanks for the changes and explanations. Looks good to me.

PR #28, opened 2021-03-03
Modify latiss_constants.py to include a sweetspot for the hologram. A…

patrickingraham:
…lso to make the platescale consistent.


tribeiro (approved):
Looks good.

PR #29, opened 2021-03-12
DM-28995 Support summit activities

tribeiro:
* Add workaround to edge condition while homing the ATDome.
    If the dome is pressing the home switch and we send a home command, it will simply register the dome as homed and won't send any event to indicate the activity is complete.
  * Add method to reset all offsets in base_tcs.
  * Add set_rem_loglevel method in RemoteGroup, that allows users to set the log level for the remotes loggers.
  * Fix "restore check" feature in prepare for flats.
  * Fix direction of PhysicalSky rotator strategy.
  * Update ATCS to support specifying rotator park position and flat field position.
    When using point_azel to slew the telescope for a safe position, use the current nasmyth position.
  * Fix setting rotFrame in xml7/8 compatibility mode.
  * Update ronchi170lpmm sweet spot.
  * Support differential ra/dec tracking in BaseTCS.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -12,7 +12,7 @@ v0.7.1
 Changes:
 
   * Add workaround to edge condition while homing the ATDome.
-    If the dome is pressing the home switch and we send a home command, it will simply register the dome as homed and won't send any event to indicate the activity is complete.
+    Now if the dome is pressing the home switch and we send a home command, it will simply register the dome as homed and won't send any event to indicate the activity is complete.
   * Add method to reset all offsets in base_tcs.
   * Add set_rem_loglevel method in RemoteGroup, that allows users to set the log level for the remotes loggers.
   * Fix "restore check" feature in prepare for flats.
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -11,6 +11,30 @@ v0.7.1
 
 Changes:
 
+  * Add workaround to edge condition while homing the ATDome.
+    If the dome is pressing the home switch and we send a home command, it will simply register the dome as homed and won't send any event to indicate the activity is complete.
+  * Add method to reset all offsets in base_tcs.
+  * Add set_rem_loglevel method in RemoteGroup, that allows users to set the log level for the remotes loggers.
+  * Fix "restore check" feature in prepare for flats.
+  * Fix direction of PhysicalSky rotator strategy.
+  * Update ATCS to support specifying rotator park position and flat field position.
+    When using point_azel to slew the telescope for a safe position, use the current nasmyth position.
+  * Fix setting rotFrame in xml7/8 compatibility mode.
+  * Update ronchi170lpmm sweet spot.
+  * Support differential ra/dec tracking in BaseTCS.
+
+Requirements:
+
+  * ts_salobj >= 5.6.0
+  * ts_xml >= 7.1.0
+  * ts_idl >= 2.0.0
+  * IDL files for all components, e.g. built with ``make_idl_files.py``
+
+v0.7.1
+======
+
+Changes:
+
   * Implement xml 7/8 compatibility.
   * Fix `add_point_data` in BaseTCS.
   * Fix timeout in opening/closing the dome.
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -135,8 +135,10 @@ class ATCS(BaseTCS):
 
         self.tel_park_el = 80.0
         self.tel_park_az = 0.0
+        self.tel_park_rot = -110.0
         self.tel_flat_el = 39.0
         self.tel_flat_az = 205.7
+        self.tel_flat_rot = -110.0
         self.tel_el_operate_pneumatics = 70.0
         self.tel_settle_time = 3.0
 
@@ -328,6 +330,7 @@ class ATCS(BaseTCS):
                 target_name="FlatField position",
                 az=self.tel_flat_az,
                 el=self.tel_flat_el,
+                rot_tel=self.tel_flat_rot,
                 wait_dome=False,
             )
 
@@ -422,6 +425,7 @@ class ATCS(BaseTCS):
             target_name="Park position",
             az=self.tel_park_az,
             el=self.tel_park_el,
+            rot_tel=self.tel_park_rot,
             wait_dome=False,
         )
         self.check.atdome = atdome_check
@@ -558,6 +562,7 @@ class ATCS(BaseTCS):
                     target_name="Park position",
                     az=self.tel_park_az,
                     el=self.tel_park_el,
+                    rot_tel=self.tel_park_rot,
                     wait_dome=False,
                 )
                 await self.stop_tracking()
@@ -817,9 +822,20 @@ class ATCS(BaseTCS):
             tel_pos = await self.next_telescope_position(timeout=self.fast_timeout)
 
             if tel_pos.elevationCalculatedAngle[-1] < self.tel_el_operate_pneumatics:
+
+                try:
+                    nasmyth = await self.rem.atmcs.tel_mount_Nasmyth_Encoders.aget(
+                        timeout=self.fast_timeout
+                    )
+                except asyncio.TimeoutError:
+                    raise RuntimeError("Cannot determine nasmyth position.")
+
+                nasmyth_angle = np.mean(nasmyth.nasmyth2CalculatedAngle)
+
                 await self.point_azel(
                     az=tel_pos.azimuthCalculatedAngle[-1],
                     el=self.tel_el_operate_pneumatics,
+                    rot_tel=nasmyth_angle,
                     wait_dome=False,
                 )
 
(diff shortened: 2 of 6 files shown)


patrickingraham (changes requested):
Looks good. Just a few minor changes requested.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -817,9 +868,20 @@ async def open_m1_cover(self):
...
             if tel_pos.elevationCalculatedAngle[-1] < self.tel_el_operate_pneumatics:
+
+                try:
+                    nasmyth = await self.rem.atmcs.tel_mount_Nasmyth_Encoders.aget(
+                        timeout=self.fast_timeout
+                    )
+                except asyncio.TimeoutError:
+                    raise RuntimeError("Cannot determine nasmyth position.")
+
+                nasmyth_angle = np.mean(nasmyth.nasmyth2CalculatedAngle)
```
patrickingraham, line 879: So this is going to cause issues if we ever want to put the instrument on the other port...
there are probably others lurking as well. Beyond the scope of this ticket but maybe we need to note somewhere that LATISS must be installed on port 2?

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -817,9 +868,20 @@ async def open_m1_cover(self):
             tel_pos = await self.next_telescope_position(timeout=self.fast_timeout)
 
             if tel_pos.elevationCalculatedAngle[-1] < self.tel_el_operate_pneumatics:
+
+                try:
```
patrickingraham, line 872: Why does this try need to be here?

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -168,6 +174,10 @@ async def slew_object(
             Default is `SkyAuto`, the rotator is positioned with respect to the
             North axis and is automacally wrapped if outside the limit. See
             `RotType` for more options.
+        dra : `float`
```
patrickingraham, line 177: I don't understand that these are used for. Seems a weird way to do non-sidereal tracking....

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -628,6 +661,25 @@ async def offset_xy(self, x, y, relative=False, persistent=False):
             )
             await self.offset_azel(az, el, relative)
 
+    async def reset_offsets(self):
+        """Reset all offsets.
```
patrickingraham, line 665: maybe say "Reset both persistent and relative offsets" 
Then say which methods to use to remove just one, or just the other....
Actually, you could add flags here such that it also has the capability to reset only persistent or only relative?

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -724,6 +724,20 @@ async def standby(self):
 
         await self.set_state(salobj.State.STANDBY)
 
+    def set_rem_loglevel(self, level):
+        """Set remotes log level.
+
+        Usefull to prevent the internal salobj warnings when read queues are
```
patrickingraham, line 730: Useful

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -724,6 +724,20 @@ async def standby(self):
...
+        filled up.
```
patrickingraham, line 731: filling up, no? it warns before full I thought...

patrickingraham (changes requested):

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -441,41 +459,56 @@ async def slew(
...
                 else self.RotFrame.TARGET,
+                epoch=epoch,
+                equinox=equinox,
+                parallax=parallax,
+                pmRA=pmRA,
+                pmDec=pmDec,
+                rv=rv,
+                dRA=dRA,
+                dDec=dDec,
+                rotMode=rot_mode if rot_mode is not None else self.RotMode.FIELD,
```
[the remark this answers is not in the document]
tribeiro: There's some documentation about how this is supposed to work as part of the package: https://ts-observatory-control.lsst.io/user-guide/tcs-user-guide-generic.html

I think it needs some updating. I can probably do that as part of DM-29197.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -315,19 +317,21 @@ async def prepare_for_flatfield(self, check=None):
 
         # Creates a copy of check so it can modify it freely to control what
         # needs to be verified at each stage of the process.
-        _check = copy.copy(self.check) if check is None else copy.copy(check)
+        check_bckup = copy.copy(self.check) if check is None else copy.copy(check)
```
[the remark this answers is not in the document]
tribeiro: Well, check is supposed to be user by users and operators to skip using the particular component. Internally it also uses it some times to skip the component, so I need to keep track of internal usages to restore them later. I probably need to document its usages in more detail.. The feature itself also needs some improvements. 

There's some documentation on the base class: https://ts-observatory-control.lsst.io/py-api/lsst.ts.observatory.control.RemoteGroup.html#remotegroup but definitely need something on the main documentation as well. I would propose to do it in DM-29197.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -635,14 +643,57 @@ async def home_dome(self):
...
+        # we issue a home command.
+        # See DM-29202 for a more permanent fix.
+
+        # Check if the dome is in zero. If it is, assume dome is homed and
+        # move forward.
+        dome_pos = await self.rem.atdome.tel_position.next(
+            flush=True, timeout=self.fast_timeout
         )
+
+        az_position = Angle(dome_pos.azimuthPosition * u.deg).wrap_at("180d")
```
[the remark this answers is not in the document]
tribeiro: So, if I grab it before sending the home command, the dome could be standing at a random position while not homed that as `azimuthPosition=0`. By grabbing it here, the dome will probably be moving already, reducing the chance of that happening.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -635,14 +643,57 @@ async def home_dome(self):
...
+                # home switch.
+                az_state = await self.rem.atdome.evt_azimuthState.next(
+                    flush=False, timeout=self.fast_timeout
+                )
+            except asyncio.TimeoutError:
+                # The event timedout, this probably means the switch was
+                # pressed. Log a warning and move on.
+                self.log.warning(
+                    "Timeout waiting for ATDome azimuthState event. This may "
+                    "indicate that the dome is already homed and pressing the home switch."
```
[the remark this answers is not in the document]
tribeiro: done

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: Yeah, I was lazy when I implemented this the first time. I'll check which one is selected and act accordingly. I am already doing this in other places as well.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: Replaced by new code.

python/lsst/ts/observatory/control/remote_group.py
tribeiro: yes

python/lsst/ts/observatory/control/remote_group.py
tribeiro: right

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: Resolved in private

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -635,14 +643,57 @@ async def home_dome(self):
...
+            az_state = await self.rem.atdome.evt_azimuthState.next(
+                flush=False, timeout=self.open_dome_shutter_time
+            )
         while az_state.homing:
             self.log.info("Dome azimuth still homing.")
             az_state = await self.rem.atdome.evt_azimuthState.next(
                 flush=False, timeout=self.open_dome_shutter_time
             )
+        else:
+            self.log.info("Dome azimuth hommed successfully.")
```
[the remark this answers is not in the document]
tribeiro: done

doc/version_history.rst
```diff
@@ -9,6 +9,30 @@ Version History
 v0.7.1
 ======
 
+Changes:
+
+  * Add workaround to edge condition while homing the ATDome.
+    If the dome is pressing the home switch and we send a home command, it will simply register the dome as homed and won't send any event to indicate the activity is complete.
```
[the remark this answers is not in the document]
tribeiro: fixed

python/lsst/ts/observatory/control/base_tcs.py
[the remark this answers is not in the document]
patrickingraham: Ok. we need a way to refer to non-docstring docs in docstrings... but that's beyond this ticket.

python/lsst/ts/observatory/control/auxtel/atcs.py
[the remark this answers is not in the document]
patrickingraham: no convinced... but not worth squabbling about :)

patrickingraham (approved):
Looks good. Maybe just add those other changes to your going to do in DM-29197 in the ticket description.
Tests are failing... but jenkins is unhappy so I can't see why...

PR #32, opened 2021-04-07
DM-28530: Add ready_to_take_data functionality in ts_observatory_control.

tribeiro:
* Add new feature to support synchronization between BaseTCS and BaseCamera.
* Implement synchronization feature in ATCS.
* Implement placeholder for synchronization feature in MTCS.
* Added documentation on how to use the synchronization between TCS and Camera.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -1527,8 +1527,8 @@ class ATCS(BaseTCS):
     async def _ready_to_take_data(self):
         """Wait until ATCS is ready to take data.
 
-        If cancelled or some condition times out or raise an exeception it will
-        set the result of the future to the exception so it propagates to
+        If cancelled or some condition times out or raises an exeception it
+        will set the result of the future to the exception so it propagates to
         whomever is waiting on it.
         """
         # Things to check
@@ -1539,7 +1539,7 @@ class ATCS(BaseTCS):
         try:
             while not ready:
                 # This loop will run until all conditions are met. Note that it
-                # may happen that one condition will be me when we run it the
+                # may happen that one condition will be met when we run it the
                 # first time and then not when we run the second time.
                 check = await asyncio.gather(
                     self.atmcs_in_position(),
@@ -1637,7 +1637,7 @@ class ATCS(BaseTCS):
     async def atmcs_in_position(self):
         """Check if atmcs is in position.
 
-        This method will try to get the next even published after the call. If
+        This method will try to get the next event published after the call. If
         it fails, it will return the last event seen. If no event was ever
         seen, it will fail with a `TimeoutError`.
 
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -135,7 +135,7 @@ class ATCS(BaseTCS):
 
         self.tel_park_el = 80.0
         self.tel_park_az = 0.0
-        self.tel_park_rot = -110.0
+        self.tel_park_rot = 0.0
         self.tel_flat_el = 39.0
         self.tel_flat_az = 205.7
         self.tel_flat_rot = -110.0
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -165,7 +165,6 @@ class ATCS(BaseTCS):
             self.rem.atmcs.evt_target.callback = self.atmcs_target_callback
 
         self.dome_az_in_position = asyncio.Event()
-        self.dome_az_in_position.clear()
 
     async def mount_AzEl_Encoders_callback(self, data):
         """Callback function to update the telescope position telemetry
--- a/python/lsst/ts/observatory/control/auxtel/atcs.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcs.py
@@ -1524,6 +1524,140 @@ class ATCS(BaseTCS):
         self.check.adome = checks.dome
         self.check.atdometrajectory = checks.dometrajectory
 
+    async def _ready_to_take_data(self):
+        """Wait until ATCS is ready to take data.
+
+        If cancelled or some condition times out or raise an exeception it will
+        set the result of the future to the exception so it propagates to
+        whomever is waiting on it.
+        """
+        # Things to check
+        # 1 - ATMCS evt_allAxesInPosition.inPosition == True
+        # 2 - All ATAOS corrections are applied
+        # 3 - ATMCS and ATDome positions are within specified range.
+        ready = False
+        try:
+            while not ready:
+                # This loop will run until all conditions are met. Note that it
+                # may happen that one condition will be me when we run it the
+                # first time and then not when we run the second time.
+                check = await asyncio.gather(
+                    self.atmcs_in_position(),
+                    self.ataos_corrections_completed(),
+                    self.dome_az_in_position.wait(),
+                )
+                self.log.debug(
+                    f"Ready to take data:: atmcs={check[0]}, ataos={check[1]}, atdome={check[2]}."
+                )
+                ready = all(check)
+        except asyncio.CancelledError as e:
+            self._ready_to_take_data_future.set_exception(e)
+        except asyncio.TimeoutError as e:
+            self._ready_to_take_data_future.set_exception(e)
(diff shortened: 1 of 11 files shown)


python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
         self.check.adome = checks.dome
         self.check.atdometrajectory = checks.dometrajectory
 
+    async def _ready_to_take_data(self):
+        """Wait until ATCS is ready to take data.
+
+        If cancelled or some condition times out or raise an exeception it will
```
wvreeven, line 1530: Change `raise` to `raises`.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
...
+        """
+        # Things to check
+        # 1 - ATMCS evt_allAxesInPosition.inPosition == True
+        # 2 - All ATAOS corrections are applied
+        # 3 - ATMCS and ATDome positions are within specified range.
+        ready = False
+        try:
+            while not ready:
+                # This loop will run until all conditions are met. Note that it
+                # may happen that one condition will be me when we run it the
```
wvreeven, line 1542: Change `me` to `met`.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
...
+                self.log.debug(
+                    f"Ready to take data:: atmcs={check[0]}, ataos={check[1]}, atdome={check[2]}."
+                )
+                ready = all(check)
+        except asyncio.CancelledError as e:
+            self._ready_to_take_data_future.set_exception(e)
+        except asyncio.TimeoutError as e:
+            self._ready_to_take_data_future.set_exception(e)
+        except Exception as e:
+            self._ready_to_take_data_future.set_exception(e)
```
wvreeven, line 1558: Is it necessary to split out these different errors like this? In the end they all call the same method so lines 1553 to 1556 are superfluous.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
...
+        ret_val = await asyncio.gather(
+            self.rem.ataos.evt_atspectrographCorrectionStarted.next(
+                flush=True, timeout=self.fast_timeout
+            ),
+            self.rem.ataos.evt_atspectrographCorrectionCompleted.next(
+                flush=True, timeout=self.fast_timeout
+            ),
+            return_exceptions=True,
+        )
+        # If they are both execeptions it probably means the ATAOS got into a
```
wvreeven, line 1598: Change `they are both` to `they both raise`.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
...
+                if not isinstance(corr, asyncio.TimeoutError)
+            ]
+
+            self.log.debug(f"Received corrections: {correction_received}.")
+            return False
+
+    async def atmcs_in_position(self):
+        """Check if atmcs is in position.
+
+        This method will try to get the next even published after the call. If
```
wvreeven, line 1640: Change `even` to `event`.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1524,6 +1524,140 @@ def unset_azel_slew_checks(self, checks):
...
+            # This means we got some data. We are not in a stable state yet.
+            correction_name = ["started", "completed"]
+            correction_received = [
+                name
+                for name, corr in zip(correction_name, ret_val)
+                if not isinstance(corr, asyncio.TimeoutError)
+            ]
+
+            self.log.debug(f"Received corrections: {correction_received}.")
+            return False
```
wvreeven, line 1635: My main concern with this approach is that `ret_val` may simply have retrieved the latest CorrectionStarted and CorrectionCompleted events so technically speaking the corrections have been applied but the code forces to wait for the next invocation where those two events time out. This seems like unnecessarily waiting for a timeout to happen. If `ret_val` contains data in the outer else block then applying the corrections may have finished if `private_sndStamp` of CorrectionStarted < `private_sndStamp` of CorrectionCompleted, right?

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -952,6 +955,20 @@ async def _offset(self, offset_cmd):
...
+        to determine the condition of the system.
+        """
+        if self._ready_to_take_data_future.done():
+            self._ready_to_take_data_future = asyncio.Future()
+            self._ready_to_take_data_task = asyncio.create_task(
+                self._ready_to_take_data()
+            )
+            return self._ready_to_take_data_future
+        else:
+            return self._ready_to_take_data_future
```
wvreeven, line 970: You can get rid of the `else` block and simply return `self._ready_to_take_data_future` in a line at the same indentation as the `if` statement.

python/lsst/ts/observatory/control/mock/atcs_mock.py
```diff
@@ -210,6 +223,31 @@ async def atptg_telemetry(self):
 
             await asyncio.sleep(1.0)
 
+    async def ataos_telemetry(self):
+
+        self.ataos.evt_correctionEnabled.set_put()
+
+        correction_times = cycle((1, 1, 2, 4, 10))
+        while self.run_telemetry_loop:
+            if self.ataos.evt_summaryState.data.summaryState == salobj.State.ENABLED:
```
wvreeven, line 232: Not entirely sure, but shouldn't this `if` be outside the `while` loop?

wvreeven (changes requested):
One concern is the way the code is forcing for a timeout while this appears to be unnecessary so please have a look at that. For the rest some minor typos and a few other possible improvements.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: so, the problem is that, if a correction was completed you cannot guarantee that there won't be another one coming just after it. The only way to be more certain about it is to make sure all corrections scheduled where completed and no new correction is done in that grace period of time.

python/lsst/ts/observatory/control/mock/atcs_mock.py
tribeiro: no, this loop is constantly running, even of the state is not enabled.

python/lsst/ts/observatory/control/auxtel/atcs.py
tribeiro: yes, the problem is that some exceptions are no longer children of `Exception`, like `CancelledError`. So I am future-proofing the code if they ever decide to make the same with `TimeoutError`.

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: good catch

tribeiro:
I am not particularly happy either.. but because the distributed nature of the system there is no good way of knowing how many corrections the ATAOS will do to the telescope position. It may be 1, 2 or more. What we know is that we will have a burst of corrections when we setup the instrument and then, they will stop, until you change the instrument setup again. These usually happens back to back at a ~1Hz frequency. So, 5s of timeout is the best compromise I could find.

python/lsst/ts/observatory/control/auxtel/atcs.py
wvreeven: OK, good idea.

python/lsst/ts/observatory/control/auxtel/atcs.py
wvreeven: OK, thanks for the explanation.

python/lsst/ts/observatory/control/mock/atcs_mock.py
wvreeven: Yeah I figured which is why I wasn't sure. Thanks for the explanation.

wvreeven (approved):
Looks good to me!

PR #33, opened 2021-04-23
DM-26457: Fix fixed rotator strategy when pointing add new support.

tribeiro:


doc/user-guide/tcs-user-guide-generic.rst
```diff
@@ -39,6 +39,59 @@ to operate the Auxiliary Telescope or;
...
+
+    radec_icrs = tcs.radec_from_azel(az=az, el=el)
+
+    ra = 5.
+    dec = -40.
+
+    azel = tcs.azel_from_radec(ra=ra, dec=dec)
+
+
+It is also possible to provide a time which the observation is intended to compute the coordinate.
```
wvreeven, line 65: "a time for which"

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -967,6 +993,117 @@ def ready_to_take_data(self):
...
+        Parameters
+        ----------
+        ra : `float`, `str` or `astropy.coordinates.Angle`
+            Target RA, either as a float (hour), a sexagesimal string
+            (HH:MM:SS.S or HH MM SS.S) coordinates or
+            `astropy.coordinates.Angle`.
+        dec : `float`, `str` or `astropy.coordinates.Angle`
+            Target Dec, either as a float (deg), a sexagesimal string
+            (DD:MM:SS.S or DD MM SS.S) coordinates or
+            `astropy.coordinates.Angle`.
```
wvreeven, line 1043: The coordinates are called az and el and not ra and dec. Please update the doc strings.

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -967,6 +993,117 @@ def ready_to_take_data(self):
...
+        time : `astropy.time.core.Time` or `None`, optional
+            The time which the coordinate trasformation is intended for. If
+            `None` (default) use current time.
+
+        Returns
+        -------
+        parallactic_angle : `astropy.coordinates.Angle`
+            Parallactic angle.
```
wvreeven, line 1090: Below you name your variable pa_angle which is computed by a function called parallactic_angle (which gets imported from the utils script). Please consider updating the doc string to refer to the pa_angle variable instead. The names of the variable and the function and even the function in which this is all coded is quite confusing :)

wvreeven (approved):
I have a few small comments that I hope you will consider and otherwise looks good.

wvreeven (changes requested):
Wait, I now see that the unit tests contain errors and they don't pass. Please fix those first.

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: bad copy and pasting... hehe

wvreeven (approved):
Thanks for the changes. Looks good!

PR #34, opened 2021-04-26
tickets/DM-29631

mareuter:


python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
...
+
+        Returns
+        -------
+        `str`
+            The set of filters available (colon delimited)
+        """
+        available_filters = await self.rem.cccamera.evt_availableFilters.aget(
+            timeout=self.fast_timeout
+        )
+        return available_filters.filterNames
```
tribeiro, line 280: Do you want to split this string and return a list of strings?

python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
         if filter is None:
```
tribeiro, line 241: Do you think this method will be most commonly called with `filter = None`? The reason I ask is that, if the most common call is to provide a filter, it may be better to do;

```
if fiter is not None:
   ...
```

python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
...
+            async with self.cmd_lock:
+                self.rem.cccamera.evt_endSetFilter.flush()
+                await self.rem.cccamera.cmd_setFilter.start(
+                    timeout=self.filter_change_timeout
+                )
+                end_setFilter = await self.rem.cccamera.evt_endSetFilter.next(
+                    flush=False, timeout=self.filter_change_timeout
+                )
+                self.log.info(f"Filter {end_setFilter.filterName} in position.")
+                return end_setFilter
```
tribeiro, line 254: Can you document the return value in the docstring? You may also want to return `None` explicitly when filter is `None`, since you do have a return value.

python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
         if filter is None:
             return
         else:
-            raise NotImplementedError("set_filter not available yet.")
+            self.rem.cccamera.cmd_setFilter.set(name=filter)
```
tribeiro, line 244: Any reason for not using `set_start` instead of `set` and then `start`?

Also, may be a good idea to provide input check. Since you have `get_available_filters`, you could make sure the filter passed is valid and raise a `RuntimeError` if it is not. You may want to skip the check if you cannot determine the filter (logging a warning, for instance).

python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
...
+
+    async def get_current_filter(self):
+        """Get the current filter.
+
+        Returns
+        -------
+        `str`
+            The filter in the light path.
+        """
+        end_setFilter = await self.rem.cccamera.evt_endSetFilter.aget(
```
tribeiro, line 264: May be a good idea to do some error handling here. You could encapsulate the await call with try/except, catch a `TimeoutError` and raise a more user-friendly message. Most users get confused with `TimeoutError` in methods like this. 

Something like this:

```
try:
    end_setFilter = await self.rem.cccamera.evt_endSetFilter.aget(...)
except asyncio.TimeoutError:
    raise RuntimeError("Could not determine current filter. Either data was never published or historical data is not working properly.")
```

python/lsst/ts/observatory/control/maintel/comcam.py
```diff
@@ -240,7 +241,43 @@ async def setup_filter(self, filter):
...
+
+    async def get_available_filters(self):
+        """Get the list of available filters.
+
+        Returns
+        -------
+        `str`
+            The set of filters available (colon delimited)
+        """
+        available_filters = await self.rem.cccamera.evt_availableFilters.aget(
```
tribeiro, line 277: Same comment as above, you may want to provide some level of error handling and produce a more user-friendly error message.

tribeiro (approved):
Overall the code looks good, I added some suggestions inline that could improve it slightly though. Mostly to add some additional error treatment and some additional high level features to the methods. 

Feel free to implement or ignore my suggestions.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: Can use `set_start`.

mareuter: Not sure what you mean by "skip the check if you cannot determine the filter". Is that in the `case filter = None`? If that's true then logging a warning for skipping the check seems odd.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: Done.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: Done.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: Done.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: List of strings sounds better. Done.

python/lsst/ts/observatory/control/maintel/comcam.py
mareuter: I think flipping the check is better since it pulls the big block of functionality up. Did look kind of weird on retrospect to see the simple "just return" line first.

tribeiro:
Feel free to merge... looks good! thanks for the updates!

PR #41, opened 2021-07-01
Tickets/dm 30729

tribeiro:


wvreeven (approved):
Very impressive mock code! I found a few typos so please correct them and also fix that flake8 error so the unit tests pass.

PR #42, opened 2021-07-06
In test_mtcs.py user `RotType.Sky` to make tests more reliable. The M…

tribeiro:
…ainTelescope rotator has a more restringing rotator angle which causes this unit tests to fail depending on the time the test runs if run with the default setting. I will make appropriate tests in the future to test both conditions.


wvreeven (approved):
LGTM

PR #43, opened 2021-07-09
Tickets/dm 30949

tribeiro:
* In ATCS update algorithm to open m1 cover.
* Add object storing and finding facility to BaseTCS.
* In ATCS add functionality to stop the monitor position loop.


wvreeven (approved):
Looks good to me.

PR #46, opened 2021-09-02
tickets/DM-30277

mareuter:
Remove the delay in ComCam image taking.


tribeiro (approved):
I assume comcam can now handle fast image taking.

mareuter:
Yes it can. At least that's what I've been told. :-)

PR #49, opened 2021-09-21
DM-31748: Implement "offline" capability in BaseTCS. find_target.

tribeiro:


python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -151,7 +163,8 @@ def object_list_get(self, name):
         """
 
         if name not in self._object_list:
-            object_table = Simbad.query_object(name)
+
+            object_table = self._query_object(name)
 
             if object_table is None:
                 raise RuntimeError(f"Could not find {name} in Simbad database.")
```
wvreeven, line 170: The `_query_object` function not only queries Simbad but also the local catalogs so this error message is not correct anymore.

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -1251,6 +1366,54 @@ def parallactic_angle(self, ra, dec, time=None):
...
+        if target is None:
+            target = await self.find_target_simbad(
+                az=az, el=el, mag_limit=mag_limit, mag_range=mag_range, radius=radius
+            )
+
+        return target
+
+    async def find_target_simbad(self, az, el, mag_limit, mag_range=2.0, radius=0.5):
         """Make a cone search in Simbad and return a target close to the
         specified position.
```
wvreeven, line 1418: According to the search criteria, only targets from the HD catalog are returned but this is not reflected in the doc string. Please clarify that here.

python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -1301,6 +1471,100 @@ async def find_target(self, az, el, mag_limit, mag_range=2.0, radius=0.5):
 
         return f"{target_main_id}".rstrip()
 
+    async def find_target_local_catalog(
+        self, az, el, mag_limit, mag_range=2.0, radius=0.5
+    ):
+        """Make a cone search in the internal catalog and return a target
+        close to the specified position.
```
wvreeven, line 1478: Contrary to the `find_target_simbad` function, this function returns any target close to the specified position (as opposed to only targets from the HD catalog). I know that only the HD catalog PD file currently is present but this could change in the future. It worries me that this distinction between the two functions exists and I suggest that this is documented well here. Also, the magnitude limit and range are given but there is no check if the range falls outside of the catalog contents (which only has stars brighter than magnitude 6).

wvreeven (changes requested):
I understand the purpose of the functions that query the local catalogs and Simbad but I am uneasy about the implementation and the accompanying documentation. Please review my comments and at least improve the documentation of the functions so their purposes are much clearer.

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: Why to you say only HD catalog starts are returned? The search criteria does not specify a catalog so targets from any catalog available in Simbad are returned.

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: So, `find_target_simbad` does not limit the search to the HD catalog, as I explained above. This one does a search in a "local catalog" loaded by the user. The catalog that ships with the module is only a test catalog, and I plan on injecting the full HD catalog and probably Gaia in the future. But I am probably not going to ship that with the module, but rather add them after the installation. 

The method does check If there are targets in the magnitude range and report an error in case there are none, see lines 1519-1521.

python/lsst/ts/observatory/control/base_tcs.py
wvreeven: Line 1454 contains `"cat = HD"` and that's added to the search criteria for Simbad.

python/lsst/ts/observatory/control/base_tcs.py
wvreeven: Thanks for the explanation regarding the magnitude range check.

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: oh wow! you are absolutely right.. I forgot I added that.. nice catch....

python/lsst/ts/observatory/control/base_tcs.py
tribeiro: I moved this check to `_query_object`, expanding the error message if the catalog is loaded and the target is not found.

wvreeven (approved):
Looks great now, thanks!

PR #53, opened 2022-01-24
DM-33250: Update control classes, scripts and scheduler to work with new image keywords

tribeiro:


mareuter (approved):
Everything reads like I would expect. Good to go.

PR #54, opened 2022-01-26
Tickets/dm 33405

tribeiro:


python/lsst/ts/observatory/control/base_tcs.py
```diff
@@ -63,16 +63,26 @@ class BaseTCS(RemoteGroup, metaclass=abc.ABCMeta):
         used to limit the resources allocated by the class by gathering some
         knowledge about the usage intention. By default allocates all
         resources.
-
+    concurrent_operation : `bool`, optional
+        Performs all tasks sequentially? If `False`, tasks like `enable` and
```
mareuter, line 67: Remove `Performs all tasks sequentially?`. I think the rest of the comment works just fine.

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -189,6 +189,9 @@ class RemoteGroup:
         used to limit the resources allocated by the class by gathering some
         knowledge about the usage intention. By default allocates all
         resources.
+    concurrent_operation : `bool`, optional
+        Performs all tasks sequentially? If `False`, tasks like `enable` and
```
mareuter, line 193: Same comment from above here.

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -766,7 +777,14 @@ async def set_state(self, state, settings=None, components=None):
             else:
                 set_ss_tasks.append(self.get_state(comp, ignore_timeout=True))
 
-        ret_val = await asyncio.gather(*set_ss_tasks, return_exceptions=True)
+        ret_val = (
+            await asyncio.gather(*set_ss_tasks, return_exceptions=True)
+            if self._concurrent_operation
+            else [
+                (await asyncio.gather(task, return_exceptions=True))[0]
```
mareuter, line 784: Can this be simplified to `await task for task in set_ss_tasks` or is it done this way to keep an exception on one task from blowing up the whole execution?

mareuter (approved):
Couple wordsmithing comments and one question, but otherwise looks fine.

python/lsst/ts/observatory/control/remote_group.py
tribeiro: yes.. the idea is to execute all tasks and, if one or more fails, return an error with that information.

PR #58, opened 2022-02-22
CAP-856: update for ts_salobj 7

r-owen:
There are two changes I am a bit worried about:
* I changed "settings" to "overrides" to match our new usage. But this changes the API, which may break some notebooks. (It is easy to fix standardscripts and externalscripts).
* There are two places that write mock ATPtg timeAndDate data with values that did not match the XML units. I updated them to write `timestamp` in TAI seconds, instead of MJD days, and `utc` as MJD days instead of hours in the day. I hope that's really what the pointing components are doing. Many of the time fields in that topic look like duplicates of each other and the descriptions could use some work.

Also: the documentation claimed that the first settingsToApply value was always used as the settingsToApply if the user didn't specify anything else. Obviously the new best default is `configurationOverride=""` (rather than any of the available overrides). Unfortunately I could not find the code that extracted the first settingsToApply value. If you can find it, please point it out so we can fix that code.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -2,14 +2,14 @@
 This configuration only affects single-package Sphinx documentation builds.
 """
 
-from documenteer.conf.pipelinespkg import *
-import lsst.ts.observatory.control
+from documenteer.conf.pipelinespkg import *  # noqa
+import lsst.ts.observatory.control  # noqa
 
 project = "ts_observatory_control"
-html_theme_options["logotext"] = project
+html_theme_options["logotext"] = project  # type: ignore # noqa
 html_title = project
 html_short_title = project
 doxylink = {}  # Avoid warning: Could not find tag file _doxygen/doxygen.tag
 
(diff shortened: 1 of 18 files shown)


tribeiro (approved):
Thank you very much for the update! 

As to handling the setting/overrides, the part of the code that fill up the `settings` (now `overrides`) with the first available settings is [here](https://github.com/lsst-ts/ts_observatory_control/blob/06b2d8dbe06c37a91cec6021ccaeeeb28551e908/python/lsst/ts/observatory/control/remote_group.py#L674-L680).

I think what we want to do now is replace `inspect_settings` with something like:

```
        overrides = dict()


        for comp in (
            self.components_attr if components_attr is None else components_attr
        ):
            if comp not in self.components_attr:
                raise RuntimeError(
                    f"{comp} not in group. Must be one of {self.components_attr}."
                )
            elif getattr(self.check, comp):
                overrides[comp] = ""
            else:
                self.log.info(f"{comp} check is disabled, skipping.")
```  

Which would fallback to using `override=""` when not provided by the user.

PR #60, opened 2022-03-29
tickets/SITCOM-256

mareuter:


tribeiro (approved):
Look good.. 

One thing to consider, maybe add a note on the version history about the origin of the change. I know we normally don't do this but it seems like a substantial enough change to justify it.

PR #64, opened 2022-05-09
DM-34732: check if a Future is done before setting its result or exception

r-owen:


python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -1763,39 +1763,25 @@ def set_azel_slew_checks(self, wait_dome):
...
-        will set the result of the future to the exception so it propagates to
-        whomever is waiting on it.
-        """
+        """Wait until ATCS is ready to take data."""
         # Things to check
         # 1 - ATMCS evt_allAxesInPosition.inPosition == True
         # 2 - All ATAOS corrections are applied
         # 3 - ATMCS and ATDome positions are within specified range.
         ready = False
-        try:
```
tribeiro, line 1777: Creating the `_ready_to_take_data` future is handled on the parent class: 

https://github.com/lsst-ts/ts_observatory_control/blob/f6bbe026a00bf465eb0d6d6696ed11ecc4ba0ee5/python/lsst/ts/observatory/control/base_tcs.py#L1204-L1217

Everything else looks fine, but this needs to be restored.

r-owen: Sorry about that. I looked at the base class and realized that it did not need both `_ready_to_take_data_future` and `_ready_to_take_data_task`. I changed the code so that BaseTcs calls `_ready_to_take_data` as an async method (which it already was in one of ATCS or MTCS but not the other). Then the subclass's implementaition need not set the state of any future. It just runs or raises. That is much simpler!

I also pushed a second commit that further simplifies ready_to_take_data (a bit) by making it an async method instead of a "synchronous function that returns a task". I think that clarifies the API a bit.

tribeiro (approved):
Alright! Thanks for the updates! I tested on ts_standardscripts and on ts_externalscripts and it all looks good!

PR #65, opened 2022-05-11
DM-34564: Improve handling of snaps

tribeiro:


python/lsst/ts/observatory/control/base_camera.py
```diff
@@ -360,6 +365,16 @@ async def take_object(
         take_imgtype: Take series of images of specified imgage type.
         setup_instrument: Set up instrument.
         expose: Low level expose method.
+
+        Notes
+        -----
+
+        Object images support two type of image sequence, a regular sequence
```
mareuter, line 372: The notes seem to imply that `n` and `n_snaps` are independent controls when I don't think that's the case. `n_snaps` feels more like a sub-sequence within the `n` observations. Not sure how to make the notes clearer here.

mareuter (approved):
Just a comment on the notes section, but otherwise this looks fine.

python/lsst/ts/observatory/control/base_camera.py
tribeiro: I did some work on the notes to try explain how it works in more detail. Hope it's better now! thanks for the review.

PR #67, opened 2022-05-14
DM-34810: Stop calling WriteTopic.write with an argument

r-owen:


python/lsst/ts/observatory/control/mock/mtcs_mock.py
```diff
@@ -600,126 +599,112 @@ async def m1m3_telemetry(self):
...
+                    xForce=np.random.rand(len(force_actuator_data.xForce)),
+                    yForce=np.random.rand(len(force_actuator_data.yForce)),
+                    zForce=np.random.rand(len(force_actuator_data.zForce)),
+                    fx=fx,
+                    fy=fy,
+                    fz=fz,
+                    mx=np.random.rand(),
+                    my=np.random.rand(),
+                    mz=np.random.rand(),
+                    forceMagnitude=math.hypot(fx, fy, fz),
```
tribeiro, line 639: NumPy is already used throughout the code.. can you use `np.hypot` instead of `math`?

tribeiro (approved):
Thanks for the update! I made an inline comment I hope you will consider, instead of using `math.hypot` use `numpy.hypot`...

PR #70, opened 2022-06-11
Tickets/dm 35108

tribeiro:
Sorry for the huge PR in advance. I decided to implement type checking throughout the package while I was moving to pyproject.toml. :) 

I tried to keep code changes to the minimum and only did a few tweaks here and there to make type checking happy. This should help you review it, I hope.


python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -653,7 +666,12 @@ async def expand_overrides(self, overrides=None):
         self.log.debug(f"Complete overrides: {complete_overrides}")
         return complete_overrides
 
-    async def set_state(self, state, overrides=None, components=None):
+    async def set_state(
+        self,
+        state: salobj.State,
+        overrides: typing.Optional[typing.Dict[str, str]] = None,
+        components: typing.Optional[typing.List[str]] = None,
+    ) -> None:
```
wvreeven, line 674: This comment refers to line 738. The signature of `traceback.format_exception` has changed in Python 3.10 leading to a failure when running MyPy. The CI build still is done with Python 3.8 and the conda build with Python 3.10 so it will be challenging to solve this.

conda/meta.yaml
```diff
@@ -7,6 +7,7 @@ source:
   path: ../
 
 build:
+  noarch: python
   script: python -m pip install --no-deps --ignore-installed .
   script_env:
```
wvreeven, line 12: I noticed that none of the scripts in the `bin` directory are used for unit testing. Still, are you sure you don't want to include any `entry points` for them here?

pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+urls = { documentation = "https://jira.lsstcorp.org/secure/Dashboard.jspa", repository = "https://github.com/lsst-ts/ts_observatory_control" }
+dynamic = [ "version" ]
+ 
+[tools.setuptools]
+package-data = { "" = "*.csv"}
+ 
+[tool.setuptools.dynamic]
+version = { attr = "setuptools_scm.get_version" }
+ 
+[tool.setuptools.packages.find]
```
wvreeven, line 19: I noticed that none of the scripts in the `bin` directory are used for unit testing. Still, are you sure you don't want to include any `entry points` for them here?

wvreeven (changes requested):
In principles this looks great. However, particularly the changed signature of `traceback.format_exception` is worrysome and requires some thinking.

PR #77, opened 2022-09-29
Tickets/dm 36357

tribeiro:


python/lsst/ts/observatory/control/script_queue.py
```diff
@@ -341,12 +341,18 @@ async def get_script_schema(self, is_standard: bool, script: str) -> str:
         await self.queue_remote.cmd_showSchema.set_start(
             isStandard=is_standard,
             path=script,
-            timeout=self.fast_timeout,
+            timeout=self.long_timeout,
```
wvreeven, line 344: This parameter is completely ignored in the code in this function. Can you remove it or make sure that it is used?

python/lsst/ts/observatory/control/script_queue.py
```diff
@@ -361,7 +367,7 @@ async def list_standard_scripts(self) -> typing.List[str]:
 
         available_scripts = await self._get_available_scripts()
 
-        return available_scripts.standard.split(",")
+        return available_scripts.standard.split(":")
```
wvreeven, line 370: Maybe extract this to a constant and use it wherever it needs to be used (for instance here and in line 383).

tests/test_script_queue.py
```diff
@@ -74,17 +74,20 @@ async def setup_types(self) -> None:
             .field_info
         )
         self.available_scripts.standard = (
-            "std_script1,std_script2,auxtel/std_script1,maintel/std_script1,"
+            "std_script1,std_script2:auxtel/std_script1,maintel/std_script1"
```
wvreeven, line 77: Please use the constant from `script_queue.py` if you chose to adopt that recommendation. Also in lines 80, 166 and 181.

wvreeven (approved):
In general this looks good. I left a few comments for your consideration.

python/lsst/ts/observatory/control/script_queue.py
tribeiro: Not sure I understand what you mean... are you suggesting the timeout is ignored?

wvreeven: Yes, that's what I was suggesting but I see now that that is not the case. Sorry, my bad.

PR #78, opened 2022-10-20
Tickets/dm 36455

tribeiro:


patrickingraham (approved):
Looks fine. I remember this as something we had to remove with the new CCS.

PR #80, opened 2023-02-21
Tickets/dm 37897

tribeiro:


edennihy (approved):
Looks good, new sweet spot is the same as the holo4_003, which agrees with what we saw on-sky.

PR #82, opened 2023-02-28
Tickets/dm 37942 - Add methods to show and assert authorization status in {{RemoteGroup}}

b1quint:
Add new utility methods in RemoteGroup to show the authorization status of a group and to assert that the user is authorized to command all CSCs in the group.

### DIFF
--- a/python/lsst/ts/observatory/control/remote_group.py
+++ b/python/lsst/ts/observatory/control/remote_group.py
@@ -1210,7 +1210,23 @@ class RemoteGroup:
 
     async def show_auth_status(self) -> None:
         """Shows the authorization status of a group"""
-        raise NotImplementedError
+
+        async def _log_auth_status(c: salobj.Remote) -> None:
+            """Helper function to log authorization status"""
+            authList = await c.evt_authList.aget()
+            self.log.info(
+                f"private_identity: {authList.private_identity}, "
+                f"authorizedUsers: {authList.authorizedUsers}, "
+                f"nonAuthorizedCSCs: {authList.nonAuthorizedCSCs}"
+            )
+
+        await asyncio.gather(
+            *[
+                _log_auth_status(getattr(self.rem, c))
+                for c in self.components_attr
+                if getattr(self.rem, c) is not None
+            ]
+        )
 
     async def assert_user_is_authorized(
         self, user: typing.Optional[str] = None
--- a/python/lsst/ts/observatory/control/remote_group.py
+++ b/python/lsst/ts/observatory/control/remote_group.py
@@ -1238,4 +1238,16 @@ class RemoteGroup:
         user : str or None, Default: None
             The user that should have the authorization to command all the CSCs in the group.
         """
-        raise NotImplementedError
+        identity = user if user else self.get_identity()
+        
+        async def _assert_user_is_authorized(c: salobj.Remote) -> None:
+            authList = await c.evt_authList.aget()
+            assert identity in authList.authorizedUsers
+        
+        await asyncio.gather(
+            *[
+                _log_auth_status(getattr(self.rem, c))
+                for c in self.components_attr
+                if getattr(self.rem, c) is not None
+            ]
+        )
--- a/python/lsst/ts/observatory/control/remote_group.py
+++ b/python/lsst/ts/observatory/control/remote_group.py
@@ -1207,3 +1207,19 @@ class RemoteGroup:
 
     async def __aexit__(self, *args: typing.Any) -> None:
         await self.close()
+
+    async def show_auth_status(self) -> None:
+        """Shows the authorization status of a group"""
+        raise NotImplementedError
+
+    async def assert_user_is_authorized(
+        self, user: typing.Optional[str] = None
+    ) -> None:
+        """Asserts that the (current) user is authorized to command all the CSCs in the current group.
+
+        Parameters
+        ----------
+        user : str or None, Default: None
+            The user that should have the authorization to command all the CSCs in the group.
+        """
+        raise NotImplementedError
--- a/tests/test_base_group.py
+++ b/tests/test_base_group.py
@@ -236,3 +236,13 @@ class TestBaseGroup(RemoteGroupTestCase, unittest.IsolatedAsyncioTestCase):
     async def test_show_auth_status(self) -> None:
         async with self.make_group(usage=Usages.DryTest):
             await self.basegroup.show_auth_status()
+
+    async def test_assert_user_is_authorized(self) -> None:
+        async with self.make_group(
+            usage=Usages.DryTest
+        ), MockAuthorizeCsc() as authorize_csc:
+            await self.basegroup.request_authorization()
+            await self.basegroup.assert_user_is_authorized()
+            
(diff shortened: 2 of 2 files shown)


python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
...
+        """
+        identity = user if user else self.get_identity()
+        
+        async def _assert_user_is_authorized(c: salobj.Remote) -> None:
+            authList = await c.evt_authList.aget()
+            assert identity in authList.authorizedUsers
+        
+        await asyncio.gather(
+            *[
+                _log_auth_status(getattr(self.rem, c))
```
tribeiro, line 1249: Are you sure this works? `_log_auth_status` is defined inside `show_auth_status` so should be out of scope here no?

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
 
     async def __aexit__(self, *args: typing.Any) -> None:
         await self.close()
+
+    async def show_auth_status(self) -> None:
+        """Shows the authorization status of a group"""
+
+        async def _log_auth_status(c: salobj.Remote) -> None:
```
tribeiro, line 1214: I think having this method defined inside the other methods might be a bit of an anti-pattern. You could use `self._aget_topic_samples_for_components` defined [here](https://github.com/lsst-ts/ts_observatory_control/blob/7c64bbd5c31761ad45d168250ca122b85601f281/python/lsst/ts/observatory/control/remote_group.py#L981) to get the values of the auth list events and then log them here.

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
...
+        """Asserts that the (current) user is authorized to command all the CSCs in the current group.
+
+        Parameters
+        ----------
+        user : str or None, Default: None
+            The user that should have the authorization to command all the CSCs in the group.
+        """
+        identity = user if user else self.get_identity()
+        
+        async def _assert_user_is_authorized(c: salobj.Remote) -> None:
```
tribeiro, line 1243: Same comment as above, I think you could use `_aget_topic_samples_for_components` and avoid having these internal methods.

tribeiro (commented):
I left a few inline comments, also, please, update the version history. I think this is a minor release, so v0.25.0

python/lsst/ts/observatory/control/remote_group.py
b1quint: I tested it on Nublado, and it works. However, I saw the comment below about being anti-pattern so I will update it.

tests/test_base_group.py
```diff
@@ -232,3 +232,17 @@ async def test_release_authorization(self) -> None:
 
             assert authorize_csc.authorized_users == f"-{self.basegroup.get_identity()}"
             assert authorize_csc.cscs_to_change == ",".join(self.basegroup.components)
+
+    async def test_show_auth_status(self) -> None:
+        async with self.make_group(usage=Usages.DryTest):
+            await self.basegroup.show_auth_status()
```
tribeiro, line 238: It would probably be a good idea to test that it logs what you expect. You can do something like this:

```
with self.assertLogs(self.basegroup.log, level=logging.INFO) as base_group_logs:
    await self.basegroup.show_auth_status()

assert "???" in base_group_logs.output
```

But replace the `"???"` with the expected text.

Also, there might be a `pytest` variant for the log assertion but I don't remember it from the top of my head.

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
...
     async def __aexit__(self, *args: typing.Any) -> None:
         await self.close()
+
+    async def show_auth_status(self) -> None:
+        """Shows the authorization status of a group"""
+
+        async def _log_auth_status(c: salobj.Remote) -> None:
+            """Helper function to log authorization status"""
+            authList = await c.evt_authList.aget()
+            self.log.info(
```
tribeiro, line 1217: I think it would be good to add the name of the component is in each log entry.

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
 
     async def __aexit__(self, *args: typing.Any) -> None:
         await self.close()
+
+    async def show_auth_status(self) -> None:
+        """Shows the authorization status of a group"""
+
+        async def _log_auth_status(c: salobj.Remote) -> None:
+            """Helper function to log authorization status"""
+            authList = await c.evt_authList.aget()
```
tribeiro, line 1216: Something I just realized, this is not specifying the timeout. This means that, if the data is not available, this will block for ever. I think we need a really short timeout (e.g. `self.fast_timeout`) but also, instead of timing out, we should format the message to display something like:

```
{component_name}: No auth list information available.
```

python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -1207,3 +1207,47 @@ async def __aenter__(self: TRemoteGroup) -> TRemoteGroup:
...
+
+        Parameters
+        ----------
+        user : str or None, Default: None
+            The user that should have the authorization to command all the CSCs in the group.
+        """
+        identity = user if user else self.get_identity()
+        
+        async def _assert_user_is_authorized(c: salobj.Remote) -> None:
+            authList = await c.evt_authList.aget()
```
tribeiro, line 1244: Make sure you specify timeouts!

tests/test_base_group.py
```diff
@@ -232,3 +232,17 @@ async def test_release_authorization(self) -> None:
...
+
+    async def test_assert_user_is_authorized(self) -> None:
+        async with self.make_group(
+            usage=Usages.DryTest
+        ), MockAuthorizeCsc() as authorize_csc:
+            await self.basegroup.request_authorization()
```
tribeiro, line 244: Maybe add a test that `self.basegroup.assert_user_is_authorized()` actually raises an exception before `self.basegroup.request_authorization()` is called?

you can do that with:

```
with pytest.raises(AssertionError):
    self.basegroup.assert_user_is_authorized()
```

tribeiro (commented):
Some additional comments.

python/lsst/ts/observatory/control/remote_group.py
b1quint: This is automatically solved when I apply your comments above. The `_aget_topic_samples_for_components` method already considers the timeout when getting the event.

python/lsst/ts/observatory/control/remote_group.py
b1quint: Same as above.

PR #85, opened 2023-04-25
SITCOM-781 - Create new move_m1m3_hp method in MTCS

b1quint:
We want to extract and (if needed) refactor the code in [hardpoint_move](https://github.com/lsst-ts/ts_m1m3supporttesting/blob/develop/MTM1M3Movements.py#L533) and write it the extracted code into the new `MTCS.move_m1m3_hp` method.

### DIFF
--- a/python/lsst/ts/observatory/control/maintel/mtcs.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcs.py
@@ -1075,6 +1075,84 @@ class MTCS(BaseTCS):
             component_name="MTRotator",
         )
 
+    async def move_m1m3_hp(self, hp_index: int, step_size: int) -> None:
+        """Move a single M1M3 hardpoint (hp) to a new position
+
+        Parameters
+        ----------
+        hp_index: `int`
+            Hardpoints index (from 0 to 5).
+        """
+        steps = [0] * 6
+        steps[hp_index] = step_size
+
+        self.log.info(f"Moving Hardpoint {hp_index} in {step_size} steps - Start")
+        await self.rem.mtm1m3.cmd_moveHardpointActuators.set_start(
+            steps=steps, timeout=self.long_timeout
+        )
+
+        # Wait a bit
+        await asyncio.sleep(1)
+
+        # Verify the commanded actuator is moving
+        hpActuatorState = await self.rem.mtm1m3.evt_hardpointActuatorState.aget()
+        assert (
+            hpActuatorState.motionState[hp_index]
+            == MTM1M3.HardpointActuatorMotionStates.STEPPING
+        )
+
+        # Timeout in case the actuator is not moving
+        timer_task = asyncio.create_task(asyncio.sleep(self.long_long_timeout))
+
+        # Wait for moving to complete or a limit switch is hit
+        hpActuatorState = await self.rem.mtm1m3.evt_hardpointActuatorState.next()
+        while (
+            hpActuatorState.motionState[hp_index]
+            == MTM1M3.HardpointActuatorMotionStates.STANDBY
+        ):
+            # Check if limit switch is hit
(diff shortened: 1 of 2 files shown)


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1075,6 +1075,81 @@ async def move_rotator(self, position: float) -> None:
...
+        steps = [0] * 6
+        steps[hp_index] = step_size
+
+        self.log.info(f"Moving Hardpoint {hp_index} in {step_size} steps - Start")
+        await self.rem.mtm1m3.cmd_moveHardpointActuators.set_start(
+            steps=steps, timeout=self.long_timeout
+        )
+
+        # Wait a bit
+        await asyncio.sleep(1)
```
tribeiro, line 1095: Why do you need to "wait a bit"?

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1075,6 +1075,81 @@ async def move_rotator(self, position: float) -> None:
...
+
+        # Verify the commanded actuator is moving
+        hpActuatorState = await self.rem.mtm1m3.evt_hardpointActuatorState.aget()
```
tribeiro, line 1098: Please! use `snake_case`...

also, I think what you want to do is call `self.rem.mtm1m3.evt_hardpointActuatorState.flush()` before you send the command than use 

```
hp_actuator_state = await self.rem.mtm1m3.evt_hardpointActuatorState.next(
    timeout=self.long_timeout,
    flush=False,
)
```

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1075,6 +1075,81 @@ async def move_rotator(self, position: float) -> None:
...
+        assert (
```
tribeiro, line 1099: Please, don't use `assert` for this. If you want to raise an exception do:

```
if hp_actuator_state.motionState[hp_index] != MTM1M3.HardpointActuatorMotionStates.STEPPING:
    raise RuntimeError("Relevant message explaining why you are raising an exception.")
```

tribeiro (commented):
Here some initial comments.. I think it might be worth looking at the other operations you do in the method and implement the same fixes I suggested so far.

python/lsst/ts/observatory/control/maintel/mtcs.py
pkubanek: This is poor man response to the fact that at least some version of hardpoint controller did not send updates before command was completed. So the script will get "hardpoint in idle" when looking for hardpoint state, decide it is and go to next steps - it will miss transition into "hardpoint stepping" which occurs when HP starts to move.

We can fix that (if not done already, but I doubt that) on CSC side, to prevent command return earlier then HP moves a bit. Or we can improve script by quering for a state, but ignoring IDLE state for a while.

Changing CSC behaviour isn't easy, as that requires changes into how command are processed. Due to simplification, commands are taken from a FIFO, one of the "pseudo" command is update/ILC pool, which runs every 20ms (50 Hz). So to fix that, I will need to introduce new command, "check HP movement completion" (or at least methods to the Command class handling that). Likely doable, but we shall start another DM-ticket to fix that.

All this pertains to the original idea "M1M3 commands are short-lived, all is reported back as events/status updates". We know it is bad.

tribeiro: ok. that's fair.. I think then, the right thing to do is give the CSC a bit of time for this to occur.. But, instead of "wait a bit", we can handle it in a more robust way.. getting `next` samples and checking that they go to the expected value before a timeout occurs..

b1quint:
Petr created a new command in m1m3 to run hardpoint tests. With that, we no longer need to implement this command (deprecated). Closing it.

PR #88, opened 2023-05-16
DM-39195: Add support for m1m3 hard point and break away tests in MTCS.

tribeiro:


pkubanek (changes requested):
Please check M13T002 for the proper way to translate the index into FA ID required for FA bump testing.

tribeiro:
> Please check M13T002 for the proper way to translate the index into FA ID required for FA bump testing.

Thanks for pointing that out! I included ts_cRIOpy as a dependency and updated the procedure to convert from id to index.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -133,11 +134,16 @@ def __init__(
...
 
-        self.m1m3_raise_timeout = 600.0  # timeout to raise m1m3, in seconds.
+        # timeout to raise m1m3, in seconds.
+        self.m1m3_raise_timeout = 600.0
 
         # Tolerance on the stability of the balance force magnitude
         self.m1m3_force_magnitude_stable_tolerance = 50.0
 
+        self._m1m3_actuator_id_index_table: dict[int, int] = dict(
+            [(fa[1], fa[0]) for fa in FATABLE]
```
pkubanek, line 144: FATABLE provides location of the indices - FATABLE_INDEX,,. to get actuator id from index, you do FATABLE[index][FATABLE_ID]., At least replacing [1] with FATABLE_xxx makes sense, creating that as function will be even better (or shall be done later in ts_cRIOpy)

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -822,6 +830,100 @@ async def _wait_for_mtm1m3_detailed_state(
...
+        Raises
+        ------
+        RuntimeError
+            If the hp test failed.
+        """
+
+        self.log.info("Checking if the hard point breakaway test has passed.")
+
+        while True:
+            hp_test_state = MTM1M3.HardpointTest(
```
pkubanek, line 851: Does that run? I bet what you need is hp_test_state = await self.rem.mtm1m3.evt_hardpointTestStatus.next(...).testState[hp-1] - drop MTM1M3HardpointTest please

python/lsst/ts/observatory/control/maintel/mtcs.py
tribeiro: Sorry, not sure I understood your comment. 

I see that I can get the actuator id from the FATABLE index but what I need to do here is the opposite, e.g. get the index for a particular actuator id.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -822,6 +830,100 @@ async def _wait_for_mtm1m3_detailed_state(
...
+            if actuator_id != bump_test_status.actuatorId:
+                raise RuntimeError(
+                    f"Expecting status from actuator {actuator_id} got {bump_test_status.actuatorId}."
+                )
+
+            primary_status = MTM1M3.BumpTest(
+                bump_test_status.primaryTest[actuator_index]
+            )
+            secondary_status = MTM1M3.BumpTest(
+                bump_test_status.secondaryTest[actuator_index]
```
pkubanek, line 900: secondaryTest size is 112. This is indexed by so called secondary index - FATABLE_SINDEX. You need to retrieve that from FATABLE, and use that to get secondary status

python/lsst/ts/observatory/control/maintel/mtcs.py
tribeiro: Oh! I see! I should use `FATABLE_INDEX` instead of hardcode the indices! nice! thanks

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -822,6 +830,100 @@ async def _wait_for_mtm1m3_detailed_state(
...
+        while True:
+            hp_test_state = MTM1M3.HardpointTest(
+                (
+                    await self.rem.mtm1m3.evt_hardpointTestStatus.next(
+                        flush=False, timeout=self.long_timeout
+                    )
+                ).testState[hp - 1]
+            )
+
+            if hp_test_state == MTM1M3.HardpointTest.FAILED:
```
pkubanek, line 859: You will not see that in tests, but will get hit when running with HW. The problem is it takes some time for hardpoints to transition from "PASSED" to "being tested state".. And "PASSED"/"FAILED" aren't clear when you issue testHardpoint. Please check current (need to pull - this is a bit new) develop branch of ts_m1m3supporttesting - the wait for bump test in M13T002 is as follow:

```
                if last_test_state is not None and last_test_state != test_state:
                    last_test_state = None
                if test_state in (MTM1M3.BumpTest.PASSED, MTM1M3.BumpTest.FAILED):
                    if last_test_state is None:
                        break
```

when last_test_state is retrieved before the test command is run, only that fully guarantees the code will wait for completion (e.g. see translation from pervious PASSED/FAILED to being tested and then to PASSED/FAILED). That shall be doe to both hardpoint and bump tests - perhaps pass last state(s) as arguments to wait function.

pkubanek (changes requested):
Please fix at least primary index/ secondary index for bump tests checks. Fixing last_state issue pointed in HP for both HP and FA tests is required as well.

python/lsst/ts/observatory/control/maintel/mtcs.py
tribeiro: Not sure I understand this... I think the problem is that in your M13T002 script you are not flushing the event before sending the command so you might be getting old samples that were issued before the command was issue. You also use a pretty short timeout (1s) whereas I am using a long timeout in waiting for the `next` event. 

Basically what I do is:

1 - flush the event queue
2 - issue the command
3 - wait for the next event (this will only return values that were published after the command was sent).

pkubanek (approved):
Looks good. Thanks.

PR #89, opened 2023-05-16
Tickets/dm 39179

edennihy:


python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -2067,6 +2067,87 @@ async def assert_m1_correction_disabled(self, message: str = "") -> None:
...
 
+    async def offset_aos_lut(
+        self,
+        z: float = 0.0,
+        x: float = 0.0,
+        y: float = 0.0,
+        u: float = 0.0,
+        v: float = 0.0,
+        m1: float = 0.0,
+        offset_telescope: bool = False,
```
tribeiro, line 2078: I think this should be `True` by default.

tests/auxtel/test_atcs.py
```diff
@@ -1848,6 +1848,50 @@ def test_set_rot_angle_alternatives_duplicated_entries(self) -> None:
...
+        await self.atcs.offset_aos_lut(**hexapod_offset)
+
+        self.atcs.rem.ataos.cmd_offset.set_start.assert_awaited_with(
+            **hexapod_offset,
+            m2=0.0,
+            timeout=self.atcs.long_timeout,
+        )
+
+        self.atcs.rem.atptg.cmd_poriginOffset.set_start.assert_not_awaited()
+
```
tribeiro, line 1894: Can you add a test that calling `self.atcs.offset_aos_lut` with positive m1 raises an exception?

tribeiro (commented):
This is really good! I only have a couple inline comments. Also, please, update the version history. This should be a new minor version `v0.27.0`.

tests/auxtel/test_atcs.py
edennihy: Sure, test added.

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -2067,6 +2067,87 @@ async def assert_m1_correction_disabled(self, message: str = "") -> None:
...
+            "z": z,
+            "u": u,
+            "v": v,
+        }
+
+        self.log.info(
+            f"Applying hexapod offset [m1, m2, x, y, z, u, v, w]: {offset_applied}."
+        )
+
+        self.rem.ataos.evt_detailedState.flush()
```
tribeiro, line 2121: Since you are not doing anything with this event, maybe don't flush it?

tribeiro (approved):
Looks good! I left one final comment.

PR #92, opened 2023-06-02
DM-39493: Fix finishing condition for the m1m3 actuator bump test.

tribeiro:


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -149,13 +149,23 @@ def __init__(
         self._m1m3_actuator_id_index_table: dict[int, int] = dict(
             [(fa[FATABLE_ID], fa[FATABLE_INDEX]) for fa in FATABLE]
         )
+        self._m1m3_actuator_index_id_table: dict[int, int] = dict(
```
pkubanek, line 152: There is M1M3FATable.actuatorIdToIndex..I can change to snake case and you can use that instead.

tribeiro: uhm.. I need to think about that.. I think that implementation will be slower than what I have here no? every time I call `M1M3FATable.actuatorIdToIndex` it will iterate over the entire array. Here I iterate once to create the dictionary and after that I get a simple dictionary lookup.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1169,16 +1164,63 @@ async def get_m1m3_bump_test_status(
             timeout=self.fast_timeout
         )
 
-        actuator_index = self.get_m1m3_actuator_index(status.actuatorId)
+        return self._extract_bump_test_status_info(status)
+
+    def _extract_bump_test_status_info(
```
pkubanek, line 1169: This can be greatly simplified when you remember the actuator ID and pass that as an argument, preferably to get_m1m3_bump_test_status. You use get_m1m3_bump_test_status only here and then in tests, always know actuator_id because that's for what you started test. Please add to get_m1m3_bump_test_status this actuator_id argument, translate that to primary and secondary indices, get bump test status and you are done. You don't need to rely on complex max operation from timestamps.

pkubanek (changes requested):
Please pass actuator_id argument to get_m1m3_bump_test_status call. You always know it. Then you can drop complex _extract_bump_test_status_info

tribeiro:
I am torn about your suggestion. In fact, I initially implemented it how suggested, adding `actuator_id` to `get_m1m3_bump_test_status`, but later I thought it would be a more elegant implementation to hide this detail from the methods interface.

I'll go with your suggestion but I am honestly not 100% sold on it.

python/lsst/ts/observatory/control/maintel/mtcs.py
pkubanek: I have plans to better organize cRIOpy bussiness... let's keep that for future.

pkubanek (approved):

PR #93, opened 2023-06-06
DM-39532: Handle m1m3 booster valve and add support for move point2point in MTCS.

tribeiro:


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1695,6 +1745,80 @@ async def _ready_to_take_data(self) -> None:
...
+
+    async def close_m1m3_booster_valve(self) -> None:
+        """Close M1M3 booster valves."""
+        if self.check.mtm1m3:
+            await self._handle_m1m3_booster_valve(open=False)
+
+    async def _handle_m1m3_booster_valve(self, open: bool) -> None:
+        """Handle opening the M1M3 booster valves"""
+
+        desired_state = "open" if open else "close"
```
wvreeven, line 1763: For almost all uses of this variable (except lines 1776 and 1802), the value should be `closed` instead of `close`.

wvreeven (approved):
Looks good to me. I left one cosmetic comment for you.

PR #95, opened 2023-07-05
DM-39741: Update MTCS methods to handle m1m3 force balance system.

tribeiro:


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -755,15 +748,24 @@ async def _handle_raise_m1m3(self) -> None:
         self.rem.mtm1m3.evt_detailedState.flush()
 
         # xml 9/10 compatibility
```
couger01, line 750: Is XML 9/10 compatibility still needed?

couger01 (approved):
LGTM. Just a question for clarification though.

python/lsst/ts/observatory/control/maintel/mtcs.py
tribeiro: you are right, it is not.. I will strip all these backwards compatibilities with a single individual ticket though.. :)

PR #100, opened 2023-08-28
First push of modified take_stuttered code

craiglagegit:


tribeiro (commented):
Instead of adding `change_focus` to the `CameraExposure` class, add the `checkpoint` method. To make typing easier I would do something like this:

```
# In CameraExposure
checkpoint: None | typing.Coroutine[None, None, None]
```

Then in `_handle_expose_shift` you would do:

```
if camera_exposure.checkpoint is not None:
    await camera_exposure.checkpoint(f"Shifting {camera_exposure.row_shift} rows.")
```

That will make the code more general.

craiglagegit:
Hi Tiago.  I've finally gotten back to this, and I've made those changes, but there is a conflict that I don't understand.  In CameraExpose I put:
checkpoint: None | typing.Coroutine[None, None, None]
like you suggested.  But checkpoint is also used to leave a comment in other cases.  So for example, in take_bias, it says:
checkpoint: typing.Optional[typing.Callable[[str], typing.Awaitable]] = None
and the linter is complaining that those types are incompatible.  I don't quite understand what I'm doing so I don't know how to fix it.  For now I have pushed the changes to both ts_observatory_control and ts_standardscripts.   Thanks.

craiglagegit:
Hi Tiago.  Me again.  I've updated the typing for "checkpoint" in camera_exposure to match what is in the other take commands, and now the linter checks pass.  I also tested in a notebook whether the "async def offset_hexapod" method is being passed, and I think it is.  The only thing failing now is updating the version history.  Do I just update that manually?  Other than that maybe it is ready to test?  I will try commenting out the row shift command and test it at the TTS.   I'd appreciate your inputs.

edennihy:
We are working on cleaning up stale PRs in our sal script repos. This feature is no longer under development, PR will be closed.

PR #105, opened 2023-10-03
Use lsst.ts.xml.tables for FATable

pkubanek:


tribeiro (approved):
Look good, by please, see my comment inline.

PR #107, opened 2023-10-11
Update ``MTCS.move_rotator`` with new ``wait_for_in_position`` option

b1quint:
The [DM-41081](https://jira.lsstcorp.org/browse/DM-41081) ticket requests a new SAL Script to send a command to the Rotator that moves it. It should also allow the script completion before the MTRotator reaches its final position. This functionality implemented here simplifies the SAL Script and avoids duplication.


doc/version_history.rst
```diff
@@ -6,6 +6,11 @@
 Version History
 ###############
 
+v0.31.1
```
tribeiro, line 9: v0.31.0 has not been released yet, please, add this to v0.31.0 below.

tribeiro (commented):
This looks good! please, just fix the version history to add this change to v0.31.0 instead.

PR #108, opened 2023-10-12
Tickets/dm 41083 - Add stop_rotator method to mtcs

edennihy:


tests/maintel/test_mtcs.py
```diff
@@ -1538,6 +1538,19 @@ async def test_move_rotator(self) -> None:
             timeout=self.mtcs.long_long_timeout, flush=False
         )
 
+    async def test_stop_rotator(self) -> None:
+        await self.mtcs.stop_rotator()
+
+        self.mtcs.rem.mtrotator.evt_controllerState.flush.assert_called()
+
+        self.mtcs.rem.mtrotator.cmd_stop.set_start.assert_called_with(
```
tribeiro, line 1546: I think this need to be `assert_awaited_with`

tests/maintel/test_mtcs.py
```diff
@@ -1538,6 +1538,19 @@ async def test_move_rotator(self) -> None:
...
+            timeout=self.mtcs.long_timeout
+        )
+
+        self.mtcs.rem.mtrotator.evt_controllerState.aget.assert_called_with(
```
tribeiro, line 1550: I think this need's to be `assert_awaited_with`.

tests/maintel/test_mtcs.py
```diff
@@ -1538,6 +1538,19 @@ async def test_move_rotator(self) -> None:
...
+            timeout=self.mtcs.long_timeout
```
tribeiro, line 1551: You probably need to add `flush=False` here as well.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1587,6 +1587,26 @@ async def move_rotator(self, position: float) -> None:
             component_name="MTRotator",
         )
 
+    async def stop_rotator(self) -> None:
+        """Stop rotator movement and wait for controller to publish Stationary
+        substate event."""
+
+        self.rem.mtrotator.evt_controllerState.flush()
+
+        await self.rem.mtrotator.cmd_stop.set_start(timeout=self.long_timeout)
```
tribeiro, line 1596: you are not passing any attribute so use `start` instead of `set_start`.

tribeiro (changes requested):
Code looks ok, however I think your unit test might need some work. 

For example, you need to make sure when you await `self.rem.mtrotator.evt_controllerState.aget` and `self.mtcs.rem.mtrotator.evt_controllerState.next` they will return something that resembles the topic. Also you might want to mock the behavior of awaiting `self.rem.mtrotator.cmd_stop.start`. I imagine something like starting a background task that would wait some time and then set the value of the controller state to `MTRotator.EnabledSubstate.STATIONARY`.


Take a look at the [MTCSAsyncMock](https://github.com/lsst-ts/ts_observatory_control/blob/develop/python/lsst/ts/observatory/control/mock/mtcs_async_mock.py) class. This is where these things should be implemented.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -406,6 +412,19 @@ async def _mtrotator_move(self, position: float) -> None:
...
+
+    async def _mtrotator_stop(self) -> None:
+        self._mtrotator_evt_controllerState.enabledSubstate = (
+            idl.enums.MTRotator.EnabledSubstate.CONSTANT_VELOCITY
+        )
+        await asyncio.sleep(0.1)
+        self._mtrotator_evt_controllerState.enabledSubstate = (
+            idl.enums.MTRotator.EnabledSubstate.STATIONARY
+        )
+        return self._mtrotator_evt_controllerState
```
tribeiro, line 426: you don't need to return anything here.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -103,6 +103,9 @@ async def setup_types(self) -> None:
             actualPosition=0.0,
         )
         self._mtrotator_evt_in_position = types.SimpleNamespace(inPosition=True)
+        self._mtrotator_evt_controllerState = types.SimpleNamespace(
```
tribeiro, line 106: Please, use `snake_case`, e.g. `_mtrotator_evt_controller_state`.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -418,6 +437,12 @@ async def mtrotator_evt_in_position_next(
         await asyncio.sleep(self.heartbeat_time)
         return self._mtrotator_evt_in_position
 
+    async def mtrotator_evt_controllerState_next(
```
tribeiro, line 440: Please, use `snake_case`, e.g. `mtrotator_evt_controller_state_next`

tribeiro (approved):
Ok, looks good! please, make sure you reorganize the commits before merging!

PR #111, opened 2023-10-20
DM-41258: Implement towncrier is ts_observatory_control.

tribeiro:


doc/news/README.rst
```diff
@@ -0,0 +1,39 @@
...
+* ``misc``: Other minor changes and/or additions
+
+An example file name would therefore look like ``DM-40534.doc.rst``.
+
+Each developer now has to create the news fragments for the changes they have made on their own branches, 
+instead of adding them to the release notes directly.
+The news fragments are then automatically integrated into the release notes by the ``towncrier`` tool.
+
+You can test how the content will be integrated into the release notes by running ``towncrier build --draft --version=v<X.XX.X>``. 
+Note that you have to run it from the root repository directory (i.e. the ``ts_standardscripts``).
```
dsanmartim, line 25: ts_standarscripts --> ts_obserbatory control

.github/workflows/changelog.yaml
```diff
@@ -2,18 +2,25 @@ name: ChangelogUpdated
 on:
```
dsanmartim, line 2: Maybe rename it to something like "fragment_creation.yaml", "new_creation.yaml" or something similar, as now it is not really a changelog action as it was before.

dsanmartim (approved):
I think it is already done but just add two minor more comments.

PR #120, opened 2023-12-12
Tickets/dm 41339

dsanmartim:
Here is the first try of the ``slew_ephem_target`` method.

### DIFF
--- a/doc/news/DM-41339.feature.rst
+++ b/doc/news/DM-41339.feature.rst
@@ -1,0 +1,1 @@
+In ``base_tcs.py`` add a ``slew_ephem_target`` method that allow both telescopes to track a target based on an ephemeris file.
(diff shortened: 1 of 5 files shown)


doc/news/DM-41339.feature.rst
```diff
@@ -0,0 +1 @@
+In ``base_tcs.py`` add a ``slew_sphem_target`` method that allow both telescopes to track a target based on an ephemeris file.
```
tribeiro, line 1: typo `slew_sphem_target` -> `slew_ephem_target`

tribeiro (approved):
As I mentioned to you on slack, I think you hit an old bug with the `MTCS._slew_to` method that still has `stop_before_slew` defaulting to `True`. Please, change it to `False` and then you should be able to assert that `stopTracking` is not awaited in the `MTCS` test as well. 

For the record, the change is here: https://github.com/lsst-ts/ts_observatory_control/blob/9484e0ec1586f64065de79d804d22352280caab3/python/lsst/ts/observatory/control/maintel/mtcs.py#L203

Everything else looks great! thanks!

PR #123, opened 2024-01-19
Ensure that compensation mode is enabled during a slew 

b1quint:


tribeiro (commented):
This looks promising.. One thing that would be good to include is a test that the thing is appropriately called.. I think you can add an `assert_compensation_mode` in [tests/maintel/test_mtcs.py](https://github.com/lsst-ts/ts_observatory_control/blob/develop/tests/maintel/test_mtcs.py) and call it in the `test_slew_icrs` test case.

I think the `assert` method would look something like this:


```
def assert_compensation_mode(self) -> None:

    for component in self.compensation_mode_components:
        if getattr(_check, component):
            assert getattr(self, f"_{component}_evt_compensation_mode").enabled
            remote = getattr(self.mtcs.rem, component)
            remote.cmd_setCompensationMode.set_start.assert_awaited_with(
                enable=1, timeout=self.mtcs.long_timeout
            )
```

tribeiro (approved):
Looks good!

PR #124, opened 2024-01-22
Tickets/dm 42402

dsanmartim:
Hello. Here is the first try of this ticket. I hope it is not too bad :)

### DIFF
--- a/python/lsst/ts/observatory/control/maintel/mtcs.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcs.py
@@ -1936,6 +1936,68 @@ class MTCS(BaseTCS):
             )
         )
 
+    def map_slew_setting_to_attribute(
+        self, setting_enum: MTM1M3.SetSlewControllerSettings
+    ) -> str:
+        """
+        Maps a SetSlewControllerSettings enum to the corresponding attribute
+        returned by the evt_slew_controller_settings.
+
+        Parameters
+        ----------
+        setting_enum : MTM1M3.SetSlewControllerSettings
+            The enum member to be mapped.
+
+        Returns
+        -------
+        str
+            The corresponding attribute name.
+        """
+        setting_to_attribute = {
+            "ACCELERATIONFORCES": "useAccelerationForces",
+            "BALANCEFORCES": "useBalanceForces",
+            "BOOSTERVALVES": "triggerBoosterValves",
+            "VELOCITYFORCES": "useVelocityForces",
+        }
+        return setting_to_attribute[setting_enum.name]
+
+    async def get_m1m3_slew_controller_settings(self) -> dict:
+        """
+        Retrieve the current M1M3 slew controller settings.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the current settings, where the keys are
+            the names used in the SetSlewControllerSettings enumeration.
+
+        Raises
+        ------
+        RuntimeError
+            If the expected attribute is not found in the event.
+        """
+        settings_event = await self.rem.mtm1m3.evt_slewControllerSettings.aget(
+            timeout=self.fast_timeout
+        )
+
+        expected_attributes = {
+            "ACCELERATIONFORCES": "useAccelerationForces",
+            "BALANCEFORCES": "useBalanceForces",
+            "BOOSTERVALVES": "triggerBoosterValves",
+            "VELOCITYFORCES": "useVelocityForces",
+        }
+
+        settings = {}
+        for key, attr in expected_attributes.items():
+            # Convert string key to enum member
+            enum_member = MTM1M3.SetSlewControllerSettings[key]
+            mapped_attr = self.map_slew_setting_to_attribute(enum_member)
+            if not hasattr(settings_event, mapped_attr):
+                raise RuntimeError(f"Expected attribute '{mapped_attr}' not found.")
+            settings[key] = getattr(settings_event, mapped_attr)
+
+        return settings
+
     async def next_m1m3_applied_balance_forces(
         self, flush: bool
     ) -> salobj.type_hints.BaseMsgType:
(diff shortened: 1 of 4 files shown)


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -2071,6 +2107,71 @@ async def wait_m1m3_settle(self) -> None:
...
+
+        Parameters
+        ----------
+        slew_setting : enum.IntEnum
+            The specific force component setting to be changed.
+        enable_slew_management : bool
+            True to enable, False to disable the specified force component
+            controlled by the slew controller.
+        """
+        setting_key = self._map_m1m3_slew_setting_to_name(slew_setting)
```
tribeiro, line 2150: I don't think you need this method. You can get the "name" of the enumeration by using:

```
slew_setting.name
```

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1928,6 +1928,42 @@ async def get_m1m3_applied_balance_forces(self) -> salobj.type_hints.BaseMsgType
...
+            "BALANCEFORCES": "useBalanceForces",
+            "BOOSTERVALVES": "triggerBoosterValves",
+            "VELOCITYFORCES": "useVelocityForces",
+        }
+
+        settings = {}
+        for key, attr in expected_attributes.items():
+            if not hasattr(settings_event, attr):
+                raise RuntimeError(
+                    f"Expected attribute '{attr}' not found in slew controller settings event."
```
tribeiro, line 1961: Maybe print what are the valid ones? something like. 

f"Expected attribute '{attr}' not found in slew controller settings event. "
f"Must be one of {expected_attributes.keys()}."

tests/maintel/test_mtcs.py
```diff
@@ -1978,3 +1980,79 @@ def assert_m1m3_booster_valve_no_m1m3(self) -> None:
...
+            desired_value=True,
+        )
+
+    async def run_set_m1m3_slew_controller_setting_test(
+        self,
+        initial_setting_value: bool,
+        desired_setting: MTM1M3.SetSlewControllerSettings,
+        desired_value: bool,
+    ) -> None:
+        # Prepare initial settings
```
tribeiro, line 2008: Consider moving these things to the `MTCSAsyncMock` class. and actually implementing a mock to the internals of `get_m1m3_slew_controller_settings`, instead of mocking the entire thing. 


Basically, in [setup_types](https://github.com/lsst-ts/ts_observatory_control/blob/a5f61067fbb3ab3fe14bc44d33e0f3420dd6fb32/python/lsst/ts/observatory/control/mock/mtcs_async_mock.py#L78C15-L78C26) you should set an attribute

```
self._evt_slew_controller_settings = types.SimpleNamespace(
            useAccelerationForces=False,
            useBalanceForces=False,
            triggerBoosterValves=False,
            useVelocityForces=False,
)
```

then define a method to retrieve the value, `mtm1m3_evt_slew_controller_settings`:

```
async def mtm1m3_evt_slew_controller_settings(
        self, *args: typing.Any, **kwargs: typing.Any
    ) -> types.SimpleNamespace:
        await asyncio.sleep(self.heartbeat_time / 4.0)
        return self._evt_slew_controller_settings
```

and set it up in the `setup_mtm1m3` method:

```
m1m3_mocks = {
          ...,
          "evt_slewControllerSettings.aget.side_effect": mtm1m3_evt_slew_controller_settings,
          ...
}
```

Then do a similar thing on the command side to mock the behavior of setting the value.

tribeiro (commented):
Looks really good... The only thing is that the tests are not following the overall idea of the testing on this module. Please, take a look at my inline comments and let me know if you need any help.

python/lsst/ts/observatory/control/maintel/mtcs.py
dsanmartim: I removed the map method here, but I needed to create a different version of it to deal with mock methods.

tests/maintel/test_mtcs.py
dsanmartim: I have it working, but not happy with everything. Please take a look and tell me if you see anything wrong or if you see things to be improved.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -2079,6 +2141,50 @@ async def wait_m1m3_settle(self) -> None:
...
+        enable_slew_management : bool
+            True to enable, False to disable the specified force component
+            controlled by the slew controller.
+        """
+        if not isinstance(slew_setting, MTM1M3.SetSlewControllerSettings):
+            raise ValueError(f"Invalid slew setting: {slew_setting}")
+
+        setting_key = slew_setting.name
+
+        if setting_key is None:
```
tribeiro, line 2164: I don't thing `setting_key` can even be `None`.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -2079,6 +2141,50 @@ async def wait_m1m3_settle(self) -> None:
...
+
+        Parameters
+        ----------
+        slew_setting : enum.IntEnum
+            The specific force component setting to be changed.
+        enable_slew_management : bool
+            True to enable, False to disable the specified force component
+            controlled by the slew controller.
+        """
+        if not isinstance(slew_setting, MTM1M3.SetSlewControllerSettings):
```
tribeiro, line 2159: I think you can remove this check here. Maybe replace the type hint of `slew_setting` to `MTM1M3.SetSlewControllerSettings`?

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -746,6 +756,47 @@ async def _mtm1m3_cmd_force_actuator_bump_test(
...
 
+    async def mtm1m3_cmd_set_slew_controller_flags(
+        self, *args: typing.Any, **kwargs: typing.Any
+    ) -> None:
+        # Extract parameters from kwargs
+        slew_setting_value = kwargs.get("slewSettings")
+        enable_slew_management = kwargs.get("enableSlewManagement")
+
+        if slew_setting_value is None or enable_slew_management is None:
+            raise ValueError(
```
tribeiro, line 767: Please, use `RuntimeError` instead.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -746,6 +756,47 @@ async def _mtm1m3_cmd_force_actuator_bump_test(
...
+                "slewSettings and enableSlewManagement parameters are required."
+            )
+
+        # Validate and convert the integer to the corresponding enumeration
+        # member name
+        try:
```
tribeiro, line 773: I think this try/except here is not really necessary. If you call `MTM1M3.SetSlewControllerSettings(slew_setting_value)` with an invalid `slew_setting_value` it is already going to raise a `ValueError` exception with enough information. So I think it is kind of duplicated effort to catch and re-raise the same exception with a different error message. 

I suggest you just call 

```
setting_enum = MTM1M3.SetSlewControllerSettings(slew_setting_value)
```

here

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -746,6 +756,47 @@ async def _mtm1m3_cmd_force_actuator_bump_test(
...
+            setting_enum = MTM1M3.SetSlewControllerSettings(slew_setting_value)
+        except ValueError:
+            raise ValueError(f"Invalid slew setting value: {slew_setting_value}")
+
+        # Mapping enum names to the expected attributes
+        setting_attr = self.mtcs.map_slew_setting_to_attribute(setting_enum)
+
+        if setting_attr is None:
+            raise ValueError(f"Invalid slew setting value: {slew_setting_value}")
```
tribeiro, line 782: please use `RuntimeError`

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -746,6 +756,47 @@ async def _mtm1m3_cmd_force_actuator_bump_test(
...
+
+        # Update the internal state to reflect the change
+        if hasattr(self._evt_slew_controller_settings, setting_attr):
+            setattr(
+                self._evt_slew_controller_settings, setting_attr, enable_slew_management
+            )
+        else:
+            raise ValueError(
```
tribeiro, line 790: Please use `RuntimeError`

tribeiro (approved):
Alright, looks good! I left a few additional comments I hope you will consider before merging.

Thanks!

python/lsst/ts/observatory/control/maintel/mtcs.py
dsanmartim: If I remove this check, test_set_m1m3_slew_controller_settings_with_invalid_setting fails.

tribeiro: Ok!

PR #128, opened 2024-04-18
Adding base_calsys to observatory control

parfa30:
Initial structure. Tried to keep it as simple as possible.
Use case thought to be something like this:

```
at_calsys = AuxTelCalSys(BaseCalsys)
at_calsys.setup_calsys(calib_type='mono')
at_calsys.configure_flat(filter='u', wavelength=350)
at_calsys.perform_flat()
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,7 +82,7 @@ pipeline {
         stage('Build documentation') {
             steps {
                 withEnv(["WHOME=${env.WORKSPACE}"]) {
-                    catchError(buildResult: 'UNSTABLE', stageResult: 'UNSTABLE') {
+                    catchError(buildResult: 'SUCCESS', stageResult: 'UNSTABLE') {
                       sh """
                           set +x
                           source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
@@ -96,7 +96,7 @@ pipeline {
         stage('Try to upload documentation') {
             steps {
                 withEnv(["WHOME=${env.WORKSPACE}"]) {
-                    catchError(buildResult: 'UNSTABLE', stageResult: 'UNSTABLE') {
+                    catchError(buildResult: 'SUCCESS', stageResult: 'UNSTABLE') {
                         sh '''
                             set +x
                             source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,12 +82,14 @@ pipeline {
         stage('Build documentation') {
             steps {
                 withEnv(["WHOME=${env.WORKSPACE}"]) {
-                    sh """
-                        set +x
-                        source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
-                        setup -r .
-                        package-docs build
-                    """
+                    catchError(buildResult: 'UNSTABLE', stageResult: 'UNSTABLE') {
+                      sh """
+                          set +x
+                          source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
+                          setup -r .
+                          package-docs build
+                      """
+                    }
                 }
             }
         }
--- a/doc/user-guide/user-guide.rst
+++ b/doc/user-guide/user-guide.rst
@@ -325,3 +325,30 @@ Generic Camera Operations
 =========================
 
 TBD
+
+Generic Calsys Operations
+=========================
+
+Each telescope will have a calibration system (Calsys) that includes all components necessary to perform flats. 
+This is accomplished by projecting light onto the calibration screen. 
+Each system has the ability to perform two types of calibrations (`calib_type`): "white light" flats, which have a broadband illumination, and "monochromatic, which have a much narrower bandwidth of illumination. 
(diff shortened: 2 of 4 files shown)


python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -0,0 +1,184 @@
...
+            if key not in self.instrument_setup_attributes:
+                raise RuntimeError(
+                    f"Invalid argument {key}."
+                    f" Must be one of {self.instrument_setup_attributes}."
+                )
+
+    @abc.abstractmethod
+    async def get_optimized_exposure_times(
+        self, calib_type: str, filter: str, wavelength: float
+    ) -> str:
```
tribeiro, line 164: The documented return type is a dict but here you say it is a str. I think this is supposed to be a dict with str key and float values right? So this should be `-> dict[str, float]`.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -0,0 +1,184 @@
...
+            If error in configuration
+
+        Notes
+        -----
+        This will include the check the configuration of the telescope, dome,
+        and camera in addition to calsys.
+        """
+        raise NotImplementedError()
+
+    async def perform_flat(self, exptime_dict: dict) -> str:
```
tribeiro, line 121: Does it really returns a `str`?

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -0,0 +1,184 @@
...
+        The dome and telescope can be moving into place concurrently.
+
+        """
+        raise NotImplementedError()
+
+    async def configure_flat(
+        self,
+        filter: str,
+        wavelength: float,
+    ) -> str:
```
tribeiro, line 98: does it really return an `str`?

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -0,0 +1,184 @@
...
+        super().__init__(
+            components=components,
+            domain=domain,
+            log=log,
+            intended_usage=intended_usage,
+        )
+
+        self.instrument_setup_attributes = set(instrument_setup_attributes)
+
+    async def setup_calsys(self, calib_type: str) -> str:
```
tribeiro, line 72: Does it really return an str?

tribeiro (commented):
This looks pretty good. I have some comments about some of the methods return types hint and also about the update to the version history. I think once those are resolved I will be ready to approve the PR.

doc/user-guide/user-guide.rst
```diff
@@ -325,3 +325,23 @@ Generic Camera Operations
 =========================
 
 TBD
+
+Generic Calsys Operations
+=========================
+
+Each telescope will have a calibration system (Calsys) that includes all components necessary to perform flats. This is accomplished by projecting light onto the calibration screen. Each system has the ability to perform two types of calibrations (`calib_type`): "white light" flats, which have a broadband illumination, and "monochromatic, which have a much narrower bandwidth of illumination. In the case of the the `maintel`, there is a separate broadband illumination source per filter band.
```
tribeiro, line 332: Please, make sure you use "semantic line break". Basically, have each phrase in a separate line.

doc/version_history.rst
```diff
@@ -8,6 +8,7 @@ Version History
 
 .. towncrier release notes start
 
+
```
tribeiro, line 11: There is an empty line here. Please, make sure no changes remain in this file

tribeiro (commented):
Alright, this looks good but I still have some small comments I hope you will address before I can approve the PR. 

In the documentation page, please, make sure you are using semantic line breaking. Basically, each phrase should be in a separate line. 

Also make sure you there's no changes left in the version_history file. It is currently adding an empty line. 

Finally, there is an error building the documentation that is caused by an issue with the build system that is being worked out by the documenteer folks. In order to work around that issue I suggest you change these lines [here](https://github.com/lsst-ts/ts_observatory_control/blob/b6519a9cf45a45a555f29cd7625b3b219896014a/Jenkinsfile#L85-L90) to be something like this:

```
                    catchError(buildResult: 'UNSTABLE', stageResult: 'UNSTABLE') {
                      sh """
                          set +x
                          source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
                          setup -r .
                          package-docs build
                      """
                    }
```

doc/user-guide/user-guide.rst
parfa30: Right, thanks!

Jenkinsfile
```diff
@@ -82,12 +82,14 @@ pipeline {
         stage('Build documentation') {
             steps {
                 withEnv(["WHOME=${env.WORKSPACE}"]) {
-                    sh """
-                        set +x
-                        source /home/saluser/.setup_dev.sh || echo "Loading env failed; continuing..."
-                        setup -r .
-                        package-docs build
-                    """
+                    catchError(buildResult: 'UNSTABLE', stageResult: 'UNSTABLE') {
```
tribeiro, line 85: Small mod to make it actually work, use

```
catchError(buildResult: 'SUCCESS', stageResult: 'UNSTABLE')
```

here and in line 99 below please.

PR #131, opened 2024-05-03
First draft of atcalsys

parfa30:
This isn't complete, but it is a first draft to get some eyes on it. There also aren't any unit tests and it hasn't passed Jenkins.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcalsys.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcalsys.py
@@ -1,0 +1,381 @@
+# This file is part of ts_observatory_control.
+#
+# Developed for the Vera Rubin Observatory Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["ATCalsys", "ATCalsysUsages"]
+
+import asyncio
+import logging
+import time
+import typing
+
+from lsst.ts import salobj
+
+from ..base_calsys import BaseCalsys
+from ..remote_group import Usages
+from ..util import CalibrationType
+from . import LATISS
+
+
+class ATCalsysUsages(Usages):
+    """ATCalsys usages definition.
+
+    Notes
+    -----
+
+    Additional usages definition:
+
+    * Setup: Enable ATWhitelight and perform warmup. Enable
+             ATMonochromator, Electrometer, and FiberSpectrographs.
+    * Configure: Adjust ATWhitelight and ATMonochromator for type of flat.
+    * TakeFlat: Take flatfield image with LATISS, Electrometer and FiberSpec
+    * DryTest: Disable CSCs and unit tests
+    """
+
+    Setup = 1 << 3
+    Configure = 1 << 4
+    TakeFlat = 1 << 5
+
+    def __iter__(self) -> typing.Iterator[int]:
+        return iter(
+            [
+                self.All,
+                self.StateTransition,
+                self.MonitorState,
+                self.MonitorHeartBeat,
+                self.Setup,
+                self.Configure,
+                self.TakeFlat,
+                self.DryTest,
+            ]
+        )
+
+
+class ATCalsys(BaseCalsys):
+    """LSST Auxiliary Telescope Calibration System.
+
+    ATCalsys encapsulates core functionality from the following CSCs:
+    ATWhiteLight, ATMonochromoter, Fiberspectrograph, Electrometer.
+
+    Parameters
+    ----------
+    domain : `lsst.ts.salobj.Domain`
+        Domain for remotes. If `None` create a domain.
+    log : `logging.Logger`
+        Optional logging class to be used for logging operations. If `None`,
+        creates a new logger. Useful to use in salobj.BaseScript and allow
+        logging in the class use the script logging.
+    intended_usage: `int`
+        Optional integer that maps to a list of intended operations. This is
+        used to limit the resources allocated by the class by gathering some
+        knowledge about the usage intention. By default allocates all
+        resources.
+    latiss: LATISS
+    """
+
+    def __init__(
+        self,
+        domain: typing.Optional[salobj.Domain] = None,
+        log: typing.Optional[logging.Logger] = None,
+        intended_usage: typing.Optional[int] = None,
+        latiss: typing.Optional[LATISS] = None,
+    ) -> None:
+
+        self.electrometer_index = 1
+        self.fiberspectrograph_index = 1
+
+        super().__init__(
+            components=[
+                "ATMonochromator",
+                "ATWhiteLight",
+                f"FiberSpectrograph:{self.fiberspectrograph_index}",
+                f"Electrometer:{self.electrometer_index}",
+            ],
+            domain=domain,
+            log=log,
+            intended_usage=intended_usage,
+        )
+
+        self.latiss = latiss
+
+        self.whitelight_startup_power = 1000
+
+        self.chiller_temp_tolerance_relative = 0.2
+        self.chiller_temperature = 20
+
+        self.exptime_dict: dict[str, float] = dict(
+            camera=0.0,
+            electrometer=0.0,
+            fiberspectrograph=0.0,
+        )
+        self.delay = 1
+
+    async def change_wavelength(self, wavelength: float) -> None:
+        """Change the ATMonochromator wavelength setting"""
+
+        await self.rem.atmonochromator.cmd_changeWavelength.set_start(
+            wavelength=wavelength, timeout=self.long_long_timeout
+        )
+
+    async def is_ready_for_flats(self) -> bool:
+        """Add doctring"""
+        # TODO (DM-44310): Implement method to check that the
+        # system is ready for flats.
+        return True
+
+    async def setup_calsys(self, calib_type: typing.Optional[CalibrationType]) -> None:
+        """Turn on the Chiller and the ATWhiteLight and wait for the
+        warmup to complete. This is independent of Calibration Type.
+        """
+        await self.start_chiller()
+        await self.wait_for_chiller_temp_within_tolerance()
+        await self.switch_lamp_on()
+        await self.wait_for_lamp_to_warm_up()
+
+    async def configure_flat(self, config_name: str) -> None:
+        """Configure the ATMonochromator according to the flat parameters
+
+        Parameters
+        ----------
+        config_name : `str`
+            name of the type of configuration you will run, which is saved
+            in the configuration.yaml files
+
+        Raises
+        ------
+        RuntimeError:
+
+        """
+        config_data = self.get_config(config_name)
+
+        await self.setup_latiss(config_data)
+
+        await self.rem.atmonochromator.cmd_updateMonochromatorSetup.set_start(
+            wavelength=config_data["wavelength"],
+            gratingType=config_data["monochromator_grating"],
+            fontExitSlitWidth=config_data["exit_slit"],
+            fontEntranceSlitWidth=config_data["entrance_slit"],
+            timeout=self.long_timeout,
+        )
+
+    async def setup_latiss(self, config_data: dict) -> None:
+        """Seteup Latiss"""
+
+        assert config_data["use_camera"]
+        assert self.latiss is not None
+        await self.latiss.setup_instrument()
+
+    async def electrometer_scan(self, duration: float) -> salobj.type_hints.BaseMsgType:
+        """Perform an electrometer scan for the specified duration.
+
+        Parameters
+        ----------
+        duration : `float`
+            Total duration of scan.
+
+        Returns
+        -------
+        RunTimeError
+
+        """
+
+        self.rem.electrometer_1.evt_largeFileObjectAvailable.flush()
+
+        try:
+            await self.electrometer.cmd_startScanDt.set_start(
+                scanDuration=duration, timeout=duration + self.long_timeout
+            )
+        except salobj.AckTimeoutError:
+            self.log.exception("Timed out waiting for the command ack. Continuing.")
+
+        # Make sure that a new lfo was created
+        lfo = await self.electrometer.evt_largeFileObjectAvailable.next(
+            timeout=self.long_timeout, flush=False
+        )
+        return lfo
+
+    async def take_fiber_spectrum(
+        self,
+        delay: float,
+        integration_time: float,
+    ) -> salobj.type_hints.BaseMsgType:
+        """Wait, then start an acquisition with the fiber spectrograph.
+
+        By default, this method will wait for `delay` seconds then start
+        an acquisition with the fiber spectrograph. Optionally the user may
+        provide a coroutine that will be awaited before the delay starts.
+
+        Parameters
+        ----------
+        delay : `float`
+            Seconds to wait before starting fiber spectrograph acquisition.
+        integration_time : `float`
+            Integration time for the fiber spectrum (seconds).
+
+        Returns
+        -------
+        RuntimeError : If exposure or return of lfo doesn't occur.
+        """
+        self.fiberspectrograph.evt_largeFileObjectAvailable.flush()
+
+        try:
+            await self.fiberspectrograph.cmd_expose.set_start(
+                integrationTime=integration_time,
+                timeout=integration_time + self.long_timeout,
+            )
+        except salobj.AckTimeoutError:
+            self.log.exception("Timed out waiting for the command ack. Continuing.")
+
+        lfo = await self.fiberspectrograph.evt_largeFileObjectAvailable.next(
+            timeout=self.long_timeout, flush=False
+        )
+        return lfo
+
+    async def perform_flat(self, exptime_dict: dict[str, float]) -> None:
+        # check dictionary
+        # Get the image number
+        # Somehow need to make sure camera is setup
+        assert self.latiss is not None
+        asyncio.gather(
+            self.electrometer_scan(duration=exptime_dict["electrometer"]),
+            self.take_fiber_spectrum(
+                delay=self.delay, integration_time=exptime_dict["fiberspectrograph"]
+            ),
+            self.latiss.take_flats(exptime=exptime_dict["camera"], nflats=1),
+        )
+
+    async def get_optimized_exposure_times(
+        self,
+        config_name: str,
+        wavelength: typing.Union[float, None],
+    ) -> dict[str, float]:
+        """Need exposure time for camera, electrometer and fiberspectrograph
+
+        TO-DO (DM-44361): implement this correctly
+        """
+        return dict(
+            camera=10.0,
+            electrometer=10.0,
+            fiberspectrograph=10.0,
+        )
+
+    async def start_chiller(self) -> None:
+        """Starts chiller to run at self.chiller_temperature"""
+
+        await self.rem.atwhitelight.cmd_setChillerTemperature.set_start(
+            temperature=self.chiller_temperature, timeout=self.long_timeout
+        )
+        await self.rem.atwhitelight.cmd_startChiller.set_start(
+            timeout=self.long_timeout
+        )
+
+    async def switch_lamp_on(self) -> None:
+        """Switches on the white light source with an output
+        power of self.whitelight_power W
+        """
+        self.rem.atwhitelight.evt_lampState.flush()
+
+        await self.rem.atwhitelight.cmd_turnLampOn.set_start(
+            power=self.whitelight_startup_power, timeout=self.long_long_timeout * 8
+        )
+
+    async def wait_for_chiller_temp_within_tolerance(self) -> None:
+        """Checks if chiller reaches set self.chiller_temperature within
+        set self.chiller_temp_tolerance_relative in
+        less than self.timeout_chiller_cool_down s.
+
+        Raises
+        ------
+        TimeOutError:
+            If the chiller doesn't reach self.chiller_temperature
+            within tolerance in self.timeout_chiller_cool_down s.
+
+        Note
+        ----
+        Taken from power_on_atcalsys.py
+        PAF: I think this can be improved.
+        """
+        start_chill_time = time.time()
+        while time.time() - start_chill_time < self.long_long_timeout * 8:
+            chiller_temps = await self.rem.atwhitelight.tel_chillerTemperatures.next(
+                flush=True, timeout=self.long_timeout
+            )
+            tel_chiller_temp = chiller_temps.supplyTemperature
+            self.log.debug(
+                f"Chiller supply temperature: {tel_chiller_temp:0.1f} deg "
+                f"[set:{chiller_temps.setTemperature} deg]."
+            )
+            if (
+                abs(chiller_temps.setTemperature - tel_chiller_temp)
+                / chiller_temps.setTemperature
+                <= self.chiller_temp_tolerance_relative
+            ):
+                chill_time = time.time() - start_chill_time
+                self.log.info(
+                    f"Chiller reached target temperature, {tel_chiller_temp:0.1f} deg "
+                    f"within tolerance, in {chill_time:0.1f} s."
+                )
+                break
+        else:
+            raise TimeoutError(
+                f"Timeout waiting after {self.long_long_timeout * 8} s "
+                f"for the chiller to chill to {chiller_temps.setTemperature} deg. "
+                f"Stayed at {tel_chiller_temp:0.1f} deg."
+            )
+
+    async def wait_for_lamp_to_warm_up(self) -> None:
+        """Confirms the white light source has warmed up and is on.
+
+        Raises:
+        ------
+        TimeOutError:
+            If the lamp fails to turn on after self.timeout_lamp_warm_up
+
+        Note
+        ----
+        Taken from power_on_atcalsys.py
+        PAF: I think this can be improved.
+        """
+        lamp_state = await self.rem.atwhitelight.evt_lampState.aget(
+            timeout=self.long_long_timeout, flush=True
+        )
+        self.log.info(
+            f"Lamp state: {self.rem.atwhitelight.LampBasicState(lamp_state.basicState)!r}."
+        )
+
+        while lamp_state.basicState != self.rem.atwhitelight.LampBasicState.ON:
+            try:
+                lamp_state = await self.rem.atwhitelight.evt_lampState.next(
+                    flush=False, timeout=self.long_long_timeout * 8
+                )
+                self.log.info(
+                    f"Lamp state: {self.rem.atwhitelight.LampBasicState(lamp_state.basicState)!r}."
+                )
+            except asyncio.TimeoutError:
+                raise RuntimeError(
+                    f"White Light Lamp failed to turn on after {self.long_long_timeout * 8} s."
+                )
+
+    @property
+    def electrometer(self) -> salobj.Remote:
+        return getattr(self.rem, f"electrometer_{self.electrometer_index}")
+
+    @property
+    def fiberspectrograph(self) -> salobj.Remote:
+        return getattr(self.rem, f"fiberspectrograph_{self.fiberspectrograph_index}")
--- a/python/lsst/ts/observatory/control/base_calsys.py
+++ b/python/lsst/ts/observatory/control/base_calsys.py
@@ -24,9 +24,11 @@ import abc
 import logging
 import typing
 
+import yaml
 from lsst.ts import salobj
 
 from .remote_group import RemoteGroup
+from .utils import CalibrationType
 
 
 class BaseCalsys(RemoteGroup, metaclass=abc.ABCMeta):
@@ -55,7 +57,6 @@ class BaseCalsys(RemoteGroup, metaclass=abc.ABCMeta):
     def __init__(
         self,
         components: typing.List[str],
-        instrument_setup_attributes: typing.List[str],
         domain: typing.Optional[salobj.Domain] = None,
         log: typing.Optional[logging.Logger] = None,
         intended_usage: typing.Optional[int] = None,
@@ -67,17 +68,20 @@ class BaseCalsys(RemoteGroup, metaclass=abc.ABCMeta):
             intended_usage=intended_usage,
         )
 
-        self.instrument_setup_attributes = set(instrument_setup_attributes)
+        self.calibration_config_file = "data/calibration_config.yaml"
 
-    async def setup_calsys(self, calib_type: str) -> None:
+    @abc.abstractmethod
+    async def setup_calsys(self, calib_type: typing.Optional[CalibrationType]) -> None:
         """Calibration instrument is prepared so that illumination source
         can be turned ON. Initial instrument settings are configured
         based on calib_type.
 
         Parameters
         ----------
-        calib_type : `str`
-            White Light or Monochromatic illumination system: ['White','Mono']
+        calib_type : CalibrationType (enum)
+            Identifies if WhiteLight or Mono. This is only relevant for the
+            MainTelescope, so it is optional. If not identified, all light
+            sources will be setup.
 
         Raises
         -------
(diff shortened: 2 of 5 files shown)


python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        self.cmd_timeout = 30
+        self.timeout_std = 20
+        self.whitelight_startup_power = 1000
+
+        # Chiller config
+        self.timeout_chiller_cool_down = 60 * 15
+        self.chiller_temp_tolerance_relative = 0.2
+        self.chiller_temperature = 20
+
+        self.filter_gratings = {
```
tribeiro, line 128: What are these two dictionaries intended for?

If they are what I think they are, I think you are trying to store calibration settings for different instrument configurations. 

One of the discussions I had with Patrick a long time ago was about how to store these in a way that would be easy to update. I think we decided that the best thing to do was to have these in configuration files that the class would read at runtime. This way, we can expand and update those as needed without the need to change the code.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        self.exptime_dict = (
+            {  # I don't know how to instantiate this so that I get dict[str: float]
+                "camera": 0.0,
+                "electrometer": 0.0,
+                "fiberspectrograph": 0.0,
+            }
+        )
+        self.delay = 1
+
+        self.ready_for_flats = False
```
tribeiro, line 145: This should not be a flag in the class. Instead we should have a method that checks that the system is in the correct state for flats. We need to make sure these classes are stateless because we can easily be in a situation where one script is used to prepare for flats and another is used to take the calibrations. Since they are completely different instances, they won't know of each other flags. 

Maybe replace this with an `is_ready_for_flats` async method that will do the check? We can probably leave the implementation for another ticket. So, I recommend creating a ticket for the implementation of this method and here you just create the method stub. Something like this:

``` 
async def is_ready_for_flats(self) -> bool:
    """ Add doctring """
    # TODO (DM-12345): Implement method to check that the system is ready for flats.
    return True 
```

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+
+    async def warmup_lightsource(self, calib_type: str) -> None:
+        """Turn on the Chiller and the ATWhiteLight and wait for the
+        warmup to complete
+
+        Parameters
+        ----------
+        calib_type : `str`
+            Identifies if White Light or Monochromatic. For the ATCalsys,
+            the light sourceis the same. However, if it is White Light
```
tribeiro, line 155: typo: sourceis -> source is

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+                "camera": 0.0,
+                "electrometer": 0.0,
+                "fiberspectrograph": 0.0,
+            }
+        )
+        self.delay = 1
+
+        self.ready_for_flats = False
+
+    async def warmup_lightsource(self, calib_type: str) -> None:
```
tribeiro, line 147: It is better to use enums instead of strings for this type of thing.

However, rather than passing a calibration and storing the value in a local attribute. Consider having different methods for each operation.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+            Identifies if White Light or Monochromatic. For the ATCalsys,
+            the light sourceis the same. However, if it is White Light
+            configure the Monochromator now.
+
+        Raises
+        ------
+        RuntimeError:
+
+        """
+        self.calib_type = calib_type
```
tribeiro, line 163: This should not be stored in a class attribute.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        try:
+            await self.start_chiller()
+            await self.wait_for_chiller_temp_within_tolerance()
+            await self.switch_lamp_on()
+            await self.wait_for_lamp_to_warm_up()
+            self.ready_for_flats = True
+        except asyncio.TimeoutError:
```
tribeiro, line 170: I think in this case it might be better to not wrap the calls in try/except. By doing this you are hiding the actual error and it will be harder to debug later.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        wavelength : `float`
+            [unit: nm]
+            Output wavelength of the monochromator
+
+        Raises
+        ------
+        RuntimeError:
+
+        """
+        assert self.ready_for_flats
```
tribeiro, line 189: Again. don't use a local variable for this. This should call the method I mentioned above that checks the system is in the desired state.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        RuntimeError:
+
+        """
+        self.calib_type = calib_type
+        try:
+            await self.start_chiller()
+            await self.wait_for_chiller_temp_within_tolerance()
+            await self.switch_lamp_on()
+            await self.wait_for_lamp_to_warm_up()
+            self.ready_for_flats = True
```
tribeiro, line 169: Remove this assignment here. As I mentioned above, this should be a check performed by the system , not a locally stored state variable.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Raises
+        ------
+        RuntimeError:
+
+        """
+        assert self.ready_for_flats
+        grating, entrance_slit, exit_slit = self._get_monochromator_setup(filter)
+
+        try:
```
tribeiro, line 192: Again, I recommend removing the try/except clause here. It is going to hide the actual problem and make it harder to debug.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+            await self.setup_monochromator(
```
tribeiro, line 193: This method is only running the command. I would remove it and just do the work here.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Returns
+        -------
+        RunTimeError
+
+        """
+
+        self.rem.electrometer_1.evt_largeFileObjectAvailable.flush()
+
+        try:
```
tribeiro, line 241: Again, I would remove the try/except here. However, this is a case where you might want to do a try/except around the `startScanDt` command and then, continue to wait for the LFOA event if the command timedout. This would certainly increase resiliency to missing acks. 

However, note that the correct exception to wait in these cases is not `asyncio.TimeoutError` but rather `salobj.AckTimeoutError`.

If you want to do something like this here I would recommend something like this:

```
try:
    await self.rem.electrometer_1.cmd_startScanDt.set_start(
                scanDuration=duration, timeout=duration + self.cmd_timeout
    )
except salobj.AckTimeoutError:
    self.log.exception("Timed out waiting for the command ack. Continuing.")

await self.rem.electrometer_1.evt_largeFileObjectAvailable.next(
                timeout=self.cmd_timeout, flush=False
            )
```

Also, please adjust the timeouts to use the ones provided by the base class as I mentioned above. 

Another thing to consider here is returning the LFOA event. This might be useful.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Returns
+        -------
+        RuntimeError : If exposure or return of lfo doesn't occur.
+        """
+        self.rem.fiberspectrograph_1.evt_largeFileObjectAvailable.flush()
+
+        await asyncio.sleep(delay)
+
+        try:
```
tribeiro, line 279: Same comment as above for the try/except.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+                timeout=integration_time + self.long_timeout,
+            )
+
+            self.rem.fiberspectrograph_1.evt_largeFileObjectAvailable.next(
+                timeout=self.long_timeout, flush=True
+            )
+        except asyncio.TimeoutError:
+            raise RuntimeError("Fiberspectrograph exposure issue.")
+
+    async def configure_camera(self, filter: str) -> None:
```
tribeiro, line 291: I don't think you need this method here. You already have an instance of the camera so you should be good to set it internally.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -182,3 +185,56 @@ async def get_optimized_exposure_times(
...
+        filter : `str`
+            Rubin Filter: ['u','g','r','i','z','y']
+        wavelength : `float`
+            (units = nm)
+            If monochromatic flats, wavelength range [350, 1200] nm
+
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
```
tribeiro, line 230: I don't think we need this method here.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -134,7 +137,7 @@ async def perform_flat(self, exptime_dict: dict) -> None:
             If there is some error in the data taking of the camera,
             electrometer or fiber spectrographs
         """
-        raise NotImplementedError()
+        await self.expose_calsys(exptime_dict)
```
tribeiro, line 140: If this method is just calling `expose_calsys` than we should get rid of one of them. There is no reason to have a method that just call another method unless you are providing backward compatibility or something like that.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -89,7 +89,9 @@ async def setup_calsys(self, calib_type: str) -> None:
         The dome and telescope can be moving into place concurrently.
 
         """
-        raise NotImplementedError()
+        await self.warmup_lightsource(calib_type)
```
tribeiro, line 92: Again, if this method is just calling another method from the class, one of them should be removed.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -89,7 +89,9 @@ async def setup_calsys(self, calib_type: str) -> None:
...
+        # Move dome into place?
```
tribeiro, line 93: No, this is not the job of the CalSys

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -89,7 +89,9 @@ async def setup_calsys(self, calib_type: str) -> None:
...
+        # Move telescope into place?
```
tribeiro, line 94: Not the job of CalSys

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -116,7 +118,8 @@ async def configure_flat(
         This will include the check the configuration of the telescope, dome,
         and camera in addition to calsys.
         """
-        raise NotImplementedError()
+        await self.configure_projector(filter, wavelength)
```
tribeiro, line 121: So, this is only calling 2 other methods, there's also no real gain in doing this. Either this is an abstract operation that need to be implemented in the Child classes or it is a concrete operation that does some real work.

tribeiro (commented):
Here is an initial round of comments. 

I see that you updated the `BaseCalSys` to include new abstract method and implemented some of the previously unimplemented method by calling these new methods. I think at this stage I would expect all the work to be in ATCalSys, either adding new concrete implementations to method of the base class or adding specialized functionality.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+            intended_usage=intended_usage,
+        )
+
+        self.entrance_slit = 0.5
+        self.exit_slit = 0.5
+
+        # White lamp config
+        self.timeout_lamp_warm_up = 60 * 15
+        self.track_lamp_warmup = 60
+        self.cmd_timeout = 30
```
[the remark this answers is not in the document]
parfa30: Done.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/base_calsys.py
parfa30: Ok. That's why I made a new method of warmup_lightsource in case we wanted to add other things.

python/lsst/ts/observatory/control/base_calsys.py
parfa30: same as above.

python/lsst/ts/observatory/control/base_calsys.py
parfa30: Done. I thought I'd add another in case we wanted to move the telescope and dome as part of this method.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Ok. all try/excepts have been removed

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Ok, done for electrometer and fiber spectrograph

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: done as above

python/lsst/ts/observatory/control/base_calsys.py
parfa30: removed expose_calsys

python/lsst/ts/observatory/control/base_calsys.py
parfa30: Ok, just have to figure out how to deal with the camera

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,421 @@
...
+        log: typing.Optional[logging.Logger] = None,
+        intended_usage: typing.Optional[int] = None,
+    ) -> None:
+
+        self.electrometer_index = 1
+        self.fiberspectrograph_index = 1
+
+        super().__init__(
+            components=[
+                "ATCamera",  # isn't this just LATISS?
```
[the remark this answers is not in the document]
parfa30: Ok. I added this. I'll likely need some help in dealing with this in teh code.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Ok. I have included all of these in the `instrument_setup_attributes` which can be saved in configuration files.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: For the ATCalsys, doesn't really matter which type of calibration it is. This is more important for the main telescope. I've removed this.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Moved it to the `instrument_setup_attributes` that can be saved in a configuration file

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+        # check dictionary
+        # Get the image number
+        # Somehow need to make sure camera is setup
+        if self.atcamera is not None:
+            asyncio.gather(
+                self.electrometer_scan(duration=exptime_dict["electrometer"]),
+                self.take_fiber_spectrum(
+                    delay=self.delay, integration_time=exptime_dict["fiberspectrograph"]
+                ),
+                self.atcamera.take_flats(exptime=exptime_dict["camera"], nflats=1),
```
tribeiro, line 282: Please, rename `self.atcamera` to `self.latiss`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+        lfo = await self.fiberspectrograph.evt_largeFileObjectAvailable.next(
+            timeout=self.long_timeout, flush=False
+        )
+        return lfo
+
+    async def perform_flat(self, exptime_dict: dict[str, float]) -> None:
+        # check dictionary
+        # Get the image number
+        # Somehow need to make sure camera is setup
+        if self.atcamera is not None:
```
tribeiro, line 276: I think that, instead of the `if` statement here you can do an

```
assert self.lattis is not None
```

and just remove `if`.. This will make sure that the method will raise an exception if the camera is not set and will make `mypy` happy. 

Also, note that I am assuming you renamed `atcamera` -> `latiss`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+                self.take_fiber_spectrum(
+                    delay=self.delay, integration_time=exptime_dict["fiberspectrograph"]
+                ),
+                self.atcamera.take_flats(exptime=exptime_dict["camera"], nflats=1),
+            )
+
+    def _get_monochromator_setup(
+        self,
+        filter: typing.Union[int, str, None],
+        calib_type: typing.Union[int, str, None],
```
tribeiro, line 288: I think `calib_type` should probably be an enumeration. Having it as an int, str or None is not a good idea. 

Please, create an enumeration called `CalibrationType` and use it instead.

You can see some examples of enumerations [here](https://github.com/lsst-ts/ts_observatory_control/blob/develop/python/lsst/ts/observatory/control/utils/enums.py). Which is probably a good place to add this enumeration as well.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+        elif calib_type == "Mono":
+            grating = 0  # Need to figure out how to make these come
+            # from a configuration self.filter_gratings[filter]
+            exit_slit = 0.5
+            entrance_slit = 0.5
+
+        return grating, entrance_slit, exit_slit
+
+    async def get_optimized_exposure_times(
+        self, calib_type: str, filter: str, wavelength: float
```
tribeiro, line 304: See my comment above about `calib_type`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+
+        Raises:
+        ------
+        TimeOutError:
+            If the lamp fails to turn on after self.timeout_lamp_warm_up
+
+        Note
+        ----
+        Taken from power_on_atcalsys.py
+        PAF: I think this can be improved.
```
tribeiro, line 390: PAF?

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+        """
+        lamp_state = await self.rem.atwhitelight.evt_lampState.aget(
+            timeout=self.long_long_timeout
+        )
+        self.log.info(
+            f"Lamp state: {self.rem.atwhitelight.LampBasicState(lamp_state.basicState)!r}."
+        )
+        while lamp_state.basicState != self.rem.atwhitelight.LampBasicState.ON:
+            try:
+                lamp_state = await self.rem.atwhitelight.evt_lampState.next(
```
tribeiro, line 400: You probably need to flush this event before getting into the while loop.

tribeiro (commented):
Alright, looking better. I still have some comments before I can approve the PR.  

Also, sorry but I confused LATISS with ATCamera in the last round of comments.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+# You should have received a copy of the GNU General Public License
+
+__all__ = ["ATCalsys", "ATCalsysUsages"]
+
+import asyncio
+import logging
+import time
+import typing
+
+import ATCamera
```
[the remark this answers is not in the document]
parfa30: Ah good! That makes more sense because it was failing

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+                self.TakeFlat,
+                self.DryTest,
+            ]
+        )
+
+
+class ATCalsys(BaseCalsys):
+    """LSST Auxiliary Telescope Calibration System.
+
+    ATCalsys encapsulates core functionality from the following CSCs: ATCamera,
```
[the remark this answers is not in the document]
parfa30: done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+                "grating",
+                "entrance_slit",
+                "exit_slit",
+            ],
+            domain=domain,
+            log=log,
+            intended_usage=intended_usage,
+        )
+
+        self.atcamera = atcamera
```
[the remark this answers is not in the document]
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+
+        self.entrance_slit = 0.5
+        self.exit_slit = 0.5
+
+        self.whitelight_startup_power = 1000
+
+        self.chiller_temp_tolerance_relative = 0.2
+        self.chiller_temperature = 20
+
+        self.wavelength = float(500)  # nm
```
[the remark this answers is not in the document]
parfa30: I have it here because you could run `configure_flat` without setting a wavelength and it will send an error because there is no self.wavelength. If you are running white light flats, it doesn't matter what self.wavelength is set to

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+
+        self.exptime_dict = (
```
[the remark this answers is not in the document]
parfa30: thanks!

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+                "electrometer": 0.0,
+                "fiberspectrograph": 0.0,
+            }
+        )
+        self.delay = 1
+
+    async def change_wavelength(self, wavelength: float) -> None:
+        """Change the ATMonochromator wavelength setting"""
+
+        self.wavelength = wavelength
```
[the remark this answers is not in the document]
parfa30: self.rem.atmonochromator.cmd_updateMonochromatorSetup.set_start() requires a wavelength. I don't think you should necessarily need to run configure_flat with a wavelength so there just needs to be something used here. I can move it from being a local variable to an optional variable for `configure_flat`

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+        self.rem.atmonochromator.cmd_changeWavelength.set_start(wavelength=wavelength)
```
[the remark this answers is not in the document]
parfa30: set time out to be `long_long_timeout`

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+
+        Raises
+        ------
+        RuntimeError:
+
+        """
+        self.check_kwargs(**kwargs)
+
+        mono_grating: typing.Union[int, float, str, None] = (
+            kwargs["grating"]
```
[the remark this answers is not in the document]
parfa30: Nice!

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+            if "grating" in kwargs and kwargs["grating"] is not None
+            else None
+        )
+        entrance_slit: typing.Union[int, float, str, None] = (
+            kwargs["entrance_slit"]
```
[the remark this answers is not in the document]
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,417 @@
...
+            if "entrance_slit" in kwargs and kwargs["entrance_slit"] is not None
+            else None
+        )
+        exit_slit: typing.Union[int, float, str, None] = (
+            kwargs["exit_slit"]
```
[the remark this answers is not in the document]
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Ok done. Also need to make sure that Latiss was configured with the correct filter, etc before this. I can do the checking in atcalsys, but should the actual configuration of the camera occur in atcalsys or in the script?

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: I'm actually going to remove `_get_monochromator_setup` as we can just call these things from a configuration file as previously suggested. I will add in an enumeration for `CalibrationType` for getting optimized exposure times and it will be useful for the Simonyi Telescope

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: Not sure if I've done this correctly

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: My initials. . .

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: added a flush a couple lines above.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,397 @@
...
+        warmup to complete
+        """
+        await self.start_chiller()
+        await self.wait_for_chiller_temp_within_tolerance()
+        await self.switch_lamp_on()
+        await self.wait_for_lamp_to_warm_up()
+
+    async def configure_flat(
+        self,
+        wavelength: typing.Union[float, None],
```
tribeiro, line 162: What you are trying to accomplish by allowing the value of the `wavelength` argument to be `None`?

The way it is right now, the user would have to call this method like:

```
await atcalsys.configure_flat(wavelength = None)
```

to have it fallback to using the "default" value of 500. I think you might be trying to provide a simple way that users could call `configure_flat` with no input at all. If that is the case I suggest you add a default value for `wavelength`. Something like:

```
async def configure_flat(self, wavelength: float = 500.0, ...) -> None:
``` 

This would allow users to do something like:

```
await atcalsys.configure_flat()
```
e.g. calling it with no argument would fallback to using 500 (or whatever value is the default).

However, I am not sure this is a good idea. You might want to stick with the original interface (at least when it comes to wavelength) and force users to always pass a value. 

This means the method definition would be something like

```
async def configure_flat(self, wavelength: float, ...) -> None:
```

And users would have to pass in a wavelength value of type float. 

However, looking at how you are later relying on `kwargs` to get additional input values, I think we may have to do some work to improve this interface in the future. I think it is fine for now to use it like this, but something to keep in mind for future improvements.

python/lsst/ts/observatory/control/utils/enums.py
```diff
@@ -138,3 +138,16 @@ class DOFName(enum.IntEnum):
...
+class CalibrationType(enum.IntEnum):
+    """Defines the different types of flat field calibrations
+
+    Whitelight: Broad spectrum light.
+
+    Mono: Monochromatic.
+
+    """
+
+    Whitelight = 1
```
tribeiro, line 152: This should probably be `WhiteLight`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,397 @@
...
+            self.electrometer_scan(duration=exptime_dict["electrometer"]),
+            self.take_fiber_spectrum(
+                delay=self.delay, integration_time=exptime_dict["fiberspectrograph"]
+            ),
+            self.latiss.take_flats(exptime=exptime_dict["camera"], nflats=1),
+        )
+
+    async def get_optimized_exposure_times(
+        self,
+        calib_type: typing.Optional[CalibrationType],
```
tribeiro, line 279: Do you really want to allow this field to be optional? I think this has to be `CalibrationType` only.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,397 @@
...
+        wavelength: typing.Optional[float],
```
tribeiro, line 280: I would also remove the `Optional` part for this field.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -0,0 +1,397 @@
...
+        **kwargs: typing.Union[int, float, str, None],
+    ) -> dict[str, float]:
+        """Need exposure time for camera, electrometer and fiberspectrograph"""
+        self.check_kwargs(**kwargs)
+
+        latiss_exptime = 10
+        electrometer_exptime = fs_exptime = latiss_exptime
+        self.exptime_dict["camera"] = float(latiss_exptime)
```
tribeiro, line 288: This is a bit confusing. Why are you updating the value of the class attribute dictionary and then returning it? 

I think just create a new dictionary and return it.

I imagine this is a temporary implementation, so just do:

```
return dict(
    camera = 10,
    electrometer = 10,
    fiberspectrograph = 10,
)
```

and maybe create a ticket for the actual implementation and add a TODO here.

python/lsst/ts/observatory/control/utils/enums.py
parfa30: done

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: correct.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: removed Optional, added option for wavelength to be None

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: done. https://rubinobs.atlassian.net/browse/DM-44361

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: now done with configuration files

tribeiro (approved):
alright! looks good.

PR #136, opened 2024-05-31
DM-44231: Support Summit Observing Weeks 19-22 of 2024

tribeiro:


edennihy (approved):
Looks good. But we may activate this feature again in the future when we finish testing ATMCS. For now it should remain disabled, so it is fine to merge.

tribeiro:
Good point.. I am going to close the PR and not merge it... thanks!

PR #139, opened 2024-06-10
initial commit to MTCalsys

parfa30:


doc/news/DM-43628.feature.rst
```diff
@@ -0,0 +1 @@
+Adding MTCalsys
```
tribeiro, line 1: Please consider:

> Adds initial implementation of MTCalsys.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+from lsst.ts.xml.enums import TunableLaser
+
+from ..base_calsys import BaseCalsys
+from ..remote_group import Usages
+from ..utils import CalibrationType
+from . import ComCam
+
+
+class MTCalsysUsages(Usages):
+    """ATCalsys usages definition.
```
tribeiro, line 38: ATCalsys -> MTCalsys

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+
+    Notes
+    -----
+
+    Additional usages definition:
+
+    * Setup: Enable ATWhitelight and perform warmup. Enable
```
tribeiro, line 45: You may want to review the messages here, they are all referring to AT.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+             ATMonochromator, Electrometer, and FiberSpectrographs.
+    * Configure: Adjust ATWhitelight and ATMonochromator for type of flat.
```
tribeiro, line 47: Still referring to AT.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+            intended_usage=intended_usage,
+        )
+
+        self.mtcamera = mtcamera
+        self.ls_select_led_location = 0.0  # um
+        self.ls_select_laser_location = 20.0  # um
+
+        self.laser_enclosure_temp = 20.0  # C
+
+        # ALl this needs to go in a config yaml file somewhere.
```
tribeiro, line 133: any reason for not putting this in a configuration file already? You could add them to the same place we add the calibration configuration.

If you rather do this later, then please, create a ticket and add a TODO here.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+            "y": 5.0,
+        }
+        self.ls_led_focus = {"u": 0.0, "g": 1.0, "r": 2.0, "i": 3.0, "z": 4.0, "y": 5.0}
+
+        self.exptime_dict: dict[str, float] = dict(
+            camera=0.0,
+            electrometer=0.0,
+            fiberspectrograph=0.0,
+        )
+        self.delay = 1
```
tribeiro, line 157: is this used somewhere?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+            "u": 0.0,
+            "g": 1.0,
+            "r": 2.0,
+            "i": 3.0,
+            "z": 4.0,
+            "y": 5.0,
+        }
+        self.ls_led_focus = {"u": 0.0, "g": 1.0, "r": 2.0, "i": 3.0, "z": 4.0, "y": 5.0}
+
+        self.exptime_dict: dict[str, float] = dict(
```
tribeiro, line 152: is this used anywhere?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+            camera=0.0,
+            electrometer=0.0,
+            fiberspectrograph=0.0,
+        )
+        self.delay = 1
+
+    def ls_laser_focus(self, wavelength: float) -> float:
```
tribeiro, line 159: In general, a function performs an action so we start it with a verb. What is this function supposed to do?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        """Function for evaluating the distance of the focus linear stage"""
```
tribeiro, line 160: You don't need to preface the docstring with "Function", but would be good to get a better understanding of what this function does. It is not clear from context here. Also, please, make sure you document the input parameters and the output value.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        return 10.0
+
+    async def change_wavelength(self, wavelength: float) -> None:
+        """Change the TunableLaser wavelength setting"""
```
tribeiro, line 164: Please, document the input parameter.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        await asyncio.gather(task_wavelength, task_focus)
+
+    async def is_ready_for_flats(self) -> bool:
+        """Add doctring"""
+        # TODO (DM-44310): Implement method to check that the
+        # system is ready for flats.
+        return True
+
+    async def setup_calsys(self, sequence_name: str) -> None:
+        """If monochromatic flats, check that laser can be enabled,
```
tribeiro, line 182: This docstring is trying to explain what the function does but it doesn't say what it is supposed to do. Maybe start with:

> Setup the calibration system.

Also, please document the input parameters.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+                distance=self.ls_select_led_location
+            )
+        else:
+            await self.linearstage_projector_select.cmd_moveAbsolute.set_start(
+                distance=self.ls_select_laser_location
+            )
+            await self.laser_setup(config_data["laser_mode"])
+            await self.rem.tunablelaser.cmd_startPropagating.set_start()
+
+    async def laser_setup(self, mode: str) -> None:
```
tribeiro, line 202: consider renaming it to `setup_laser`.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        """Is the laser on? what is the temp like? thermal control on?"""
```
tribeiro, line 203: This docustring is a bit weird, can you please rephrase it? also, please document the input parameter.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        # check the thermal system is running
+        # check set temp
+        # check actual temp and determine if good enough
+        # check the interlockState
+
+        if mode == TunableLaser.Mode.CONTINUOUS:
+            await self.rem.tunablelaser.cmd_setContinuousMode.set_start()
```
tribeiro, line 210: use `start` instead of `set_start`. Also, make sure you pass in a timeout.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        elif (mode == TunableLaser.Mode.BURST) | (mode == TunableLaser.Mode.TRIGGER):
```
tribeiro, line 211: might be better to use a set instead of `or`. something like this: 

```
elif mod in {TunableLaser.Mode.BURST, TunableLaser.Mode.TRIGGER}:
```

Also, what happens if mode is not any of these options? do you want to raise an exception?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+            await self.rem.tunablelaser.cmd_setBurstMode.set_start()
```
tribeiro, line 212: Same as above, use `start` instead of `set_start` and pass a timeout

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+
+        calibration_type = getattr(CalibrationType, str(config_data["calib_type"]))
+        filter_name = config_data[
+            "filter"
+        ]  # not sure if this is an enumeration somewhere?
+
+        task_setup_camera = (
+            self.mtcamera.setup_instrument(
+                filter=config_data["atspec_filter"],
+                grating=config_data["atspec_grating"],
```
tribeiro, line 238: no grating in comcam and lsstcam

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        config_data = self.get_calibration_configuration(config_name)
+
+        calibration_type = getattr(CalibrationType, str(config_data["calib_type"]))
+        filter_name = config_data[
+            "filter"
+        ]  # not sure if this is an enumeration somewhere?
+
+        task_setup_camera = (
+            self.mtcamera.setup_instrument(
+                filter=config_data["atspec_filter"],
```
tribeiro, line 237: This is probably not atspec_filter

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        config_data = self.get_calibration_configuration(sequence_name)
+
+        calibration_type = getattr(CalibrationType, str(config_data["calib_type"]))
+        if calibration_type == CalibrationType.WhiteLight:
+            calibration_wavelengths = np.array([float(config_data["wavelength"])])
+        else:
+            wavelength = float(config_data["wavelength"])
+            wavelength_width = float(
+                config_data["wavelength_width"]
+            )  # This doesn't make any sense
```
tribeiro, line 306: What doesn't make any sense?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,567 @@
...
+        if self.mtcamera is None and config_data["use_camera"]:
+            raise RuntimeError(
+                f"MTCamera is not defined but {config_name} requires it. "
+                "Make sure you are instantiating ComCam and passing it to MTCalsys."
+            )
+
+        if calibration_type == CalibrationType.WhiteLight:
+            leds_ = self.led_per_filter[filter_name]
+            task_select_wavelength = (
+                self.linearstage_led_select.cmd_moveAbsolute.set_start(
```
tribeiro, line 252: shouldn't this use the change_wavelength function instead?

tribeiro (commented):
Here is an initial round of comments. I stopped the review in the `run_calibration_sequence` method. I see there is a comment there about something not making sense so feel like this is a good stopping point.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: Mostly just wanted to check with you first about the location and format you'd like to see for something like this.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: You have put the exposure times in the configuration file instead of this dictionary. I had originally thought the exposure times would be generated within the code.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: I updated it in the function, but it tells us where the linear stage is supposed to move to maintain focus in the Laser projector

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: It's very different with the LED - you move a stage for a given filter. I don't think putting those two things in the same function makes sense. But perhaps I change the name of `change_wavelength` to `change_laser_wavelength`

tribeiro: Do you mean the MTCalsys has 2 ways to change the wavelength? we can the laser wavelength or the LEDs? 

If that is the case, we have 2 options:

1. We can pick one (probably the most commonly used one) to be encapsulated by the `change_wavelength` method and create another method to encapsulate the other operation (e.g. `change_laser_wavelength`).

or

2. We introduce a parameter to the `change_wavelength` method that would allow us to pass what type of "operation" we want to do. It could be something like this"

```
async def change_wavelength(self, wavelength: float, type: CalibrationType) -> None:
    ...
```

I am assuming above that you would be able to determine what do to from the calibration type, but this could be a new enumeration. We would also have to figure out what to do for ATCalsys, since it only has 1 mode. Also, ideally we would make this optional (e.g. `type: CalibrationType | None = None`), and fallback to a "standard" mode for MTCalcys.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: It's fine. Just the use of the word width makes me think of step size rather than the width of a filter/band.

parfa30:
I have decided to implement the optimized_exposure_times in another ticket because it will be for both `mtcalsys.py` and `atcalsys.py`. https://rubinobs.atlassian.net/issues/DM-44361

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -159,6 +159,32 @@ async def run_calibration_sequence(
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
```
tribeiro, line 162: If this is going to be an abstract method, you will also need to add it to `ATCalsys`. You can leave it not implemented for now, just do:

```
    async def calculate_optimized_exposure_times(
        self, sequence_name: str
    ) -> dict[str, float]:
    """ <Docstring>
    """
    raise NotImplementedError()
```

python/lsst/ts/observatory/control/maintel/mtcalsys.py
tribeiro: ping

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,644 @@
...
+    log : `logging.Logger`
+        Optional logging class to be used for logging operations. If `None`,
+        creates a new logger. Useful to use in salobj.BaseScript and allow
+        logging in the class use the script logging.
+    intended_usage: `int`
+        Optional integer that maps to a list of intended operations. This is
+        used to limit the resources allocated by the class by gathering some
+        knowledge about the usage intention. By default allocates all
+        resources.
+    mtcamera: ComCam or LSSTCam
```
tribeiro, line 93: Please add docstring.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,644 @@
...
+            log=log,
+            intended_usage=intended_usage,
+        )
+
+        self.mtcamera = mtcamera
+        self.ls_select_led_location = 9.96  # mm
+        self.ls_select_laser_location = 79.96  # mm
+        self.led_rest_position = 120.0  # mm
+
+        self.led_projector_config_filename = "../data/mtledprojector.yaml"
```
tribeiro, line 134: If you follow the naming convention established in the BaseCalsys, you don't need to do this. Please, see how this is done on BaseCalsys https://github.com/lsst-ts/ts_observatory_control/blob/16125387aa72c293774925c50311e0a7f3565db8/python/lsst/ts/observatory/control/base_calsys.py#L162

you should not have to do this here.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,644 @@
...
+            wavelength=wavelength, timeout=self.long_long_timeout
+        )
+        task_focus = self.linearstage_laser_focus.cmd_moveAbsolute.set_start(
+            distance=self.calculate_laser_focus_location(wavelength)
+        )
+
+        await asyncio.gather(task_wavelength, task_focus)
+
+    async def is_ready_for_flats(self) -> bool:
+        """Add doctring"""
```
tribeiro, line 184: Please add docstring.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,644 @@
...
+            await self.linearstage_projector_select.cmd_moveAbsolute.set_start(
+                distance=self.ls_select_laser_location, timeout=self.long_timeout
+            )
+            await self.setup_laser(config_data["laser_mode"])
+            await self.rem.tunablelaser.cmd_startPropagating.start(
+                timeout=self.long_long_timeout
+            )
+
+    async def setup_laser(self, mode: str) -> None:
+        """Perform all steps for preparing the laser for monchromatic flats.
```
tribeiro, line 228: typo:

monchromatic -> monochromatic

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,644 @@
...
+                mtcamera_exposure_info=mtcamera_exposure_info,
+            )
+
+            calibration_summary["steps"].append(step)
+        return calibration_summary
+
+    async def calculate_optimized_exposure_times(
+        self, sequence_name: str
+    ) -> dict[str, float]:
+        """Calculates the exposure times for the electrometer and
```
tribeiro, line 413: Please, see my notes on the base class about the docstring.

python/lsst/ts/observatory/control/data/mtledprojector.yaml
tribeiro, line 1: please rename this file `mtcalsys.yaml` to match the expected value by the base class.

tribeiro (commented):
I left a few additional inline comments. It is looking pretty good.

python/lsst/ts/observatory/control/data/mtledprojector.yaml
parfa30: So, this is slightly different than what we want for `mtcalsys.yaml` I think. That file should be used for different configurations of the whole system and this is just for the LED setups. Or do you want to put it all in one file?

tribeiro: But why would we need a different configuration to setup the LED?

parfa30: The items in this configuration file are more elements of the design rather than configuration choices. I see them as quite different. However, we could put it all together in one file if that is easier.

parfa30:
I will likely change all the calculate_optimized_exposure_time output to np.arrays, like we did with ATCalsys, but I will leave it for another ticket.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,756 @@
...
+        ----------
+        wavelength : `float`
+            wavelength of the laser in nm
+        """
+
+        task_wavelength = self.rem.tunablelaser.cmd_changeWavelength.set_start(
+            wavelength=wavelength, timeout=self.long_long_timeout
+        )
+        task_focus = self.linearstage_laser_focus.cmd_moveAbsolute.set_start(
+            distance=self.calculate_laser_focus_location(wavelength)
```
tribeiro, line 198: Please, add a timeout.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -0,0 +1,756 @@
...
+
+    async def setup_laser(self, mode: str) -> None:
+        """Perform all steps for preparing the laser for monochromatic flats.
+        This includes confirming that the thermal system is
+        turned on and set at the right temperature. It also checks
+        the interlockState to confirm it's ready to propagate.
+
+        Parameters
+        ----------
+        mode : `str`
```
tribeiro, line 255: This should be an enumeration no?

tribeiro (approved):
Looks good, I left a few in line comments but I think we can merge it how it is and we make some improvements later.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: Glad you pointed this out. There are two different enums: one for the XML and one in the TunableLaser. I have to use the one from the XML, right?

tribeiro: yes

PR #141, opened 2024-06-10
Tickets/dm 44670

elanaku:
ATCalSys updated so that (1) the filter scans have the correct wavelength range, (2) that setup_electrometers takes inputs from the config file about the electrometer mode, range, and integration time, and (3) there is only one electrometer scan per exposure.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcalsys.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcalsys.py
@@ -196,9 +196,16 @@ class ATCalsys(BaseCalsys):
             else utils.make_done_future()
         )
 
+        task_setup_electrometer = self.setup_electrometers(
+            mode=str(config_data["electrometer_mode"]),
+            range=float(config_data["electrometer_range"]),
+            integration_time=float(config_data["electrometer_integration_time"]),
+        )
+
         await asyncio.gather(
             task_setup_monochromator,
             task_setup_latiss,
+            task_setup_electrometer,
         )
 
     async def run_calibration_sequence(
--- a/python/lsst/ts/observatory/control/auxtel/atcalsys.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcalsys.py
@@ -233,7 +233,7 @@ class ATCalsys(BaseCalsys):
 
         calibration_type = getattr(CalibrationType, str(config_data["calib_type"]))
         if calibration_type == CalibrationType.WhiteLight:
-            calibration_wavelenghts = np.array([float(config_data["wavelength"])])
+            calibration_wavelengths = np.array([float(config_data["wavelength"])])
         else:
             wavelength = float(config_data["wavelength"])
             wavelength_width = float(config_data["wavelength_width"])
@@ -241,11 +241,13 @@ class ATCalsys(BaseCalsys):
             wavelength_start = wavelength - wavelength_width / 2.0
             wavelength_end = wavelength + wavelength_width / 2.0
 
-            calibration_wavelenghts = np.arange(
-                wavelength_start, wavelength_end, wavelength_resolution
+            calibration_wavelengths = np.arange(
+                wavelength_start,
+                wavelength_end + wavelength_resolution,
+                wavelength_resolution,
             )
 
-        for wavelength in calibration_wavelenghts:
+        for wavelength in calibration_wavelengths:
             self.log.debug(
                 f"Performing {calibration_type.name} calibration with {wavelength=}."
             )
--- a/python/lsst/ts/observatory/control/base_calsys.py
+++ b/python/lsst/ts/observatory/control/base_calsys.py
@@ -28,6 +28,7 @@ import typing
 
 import yaml
 from lsst.ts import salobj
+from lsst.ts.xml.enums.Electrometer import UnitToRead
 
 from .remote_group import RemoteGroup
 from .utils import get_data_path
@@ -72,13 +73,39 @@ class BaseCalsys(RemoteGroup, metaclass=abc.ABCMeta):
 
         self.calibration_config: dict[str, dict[str, typing.Any]] = dict()
 
-    async def setup_electrometers(self) -> None:
-        """Setup all electrometers."""
+    async def setup_electrometers(
+        self, mode: str, range: float, integration_time: float
+    ) -> None:
+        """Setup all electrometers.
+
+        Parameters
+        ----------
+        mode : `str`
+            Electrometer measurement mode.
+        range : `float`
+            Electrometer measurement range. -1 for autorange.
+        integration_time : `float`
+            Electrometer measurement range.
+        """
+        electrometer_mode = getattr(UnitToRead, mode).value
+
         for electrometer in [
             getattr(self.rem, component_name)
             for component_name in self.components_attr
             if "electrometer" in component_name
         ]:
+            await electrometer.cmd_setMode.set_start(
+                mode=electrometer_mode,
+                timeout=self.long_timeout,
+            )
+            await electrometer.cmd_setRange.set_start(
+                setRange=range,
+                timeout=self.long_timeout,
+            )
+            await electrometer.cmd_setIntegrationTime.set_start(
+                intTime=integration_time,
+                timeout=self.long_timeout,
+            )
             await electrometer.cmd_performZeroCalib.start(timeout=self.long_timeout)
             await electrometer.cmd_setDigitalFilter.set_start(
                 activateFilter=False,
--- a/doc/news/DM-44670.bugfix.rst
+++ b/doc/news/DM-44670.bugfix.rst
@@ -1,0 +1,1 @@
+Update ATCalSys so that the filter scans have the correct wavelength range.
--- a/doc/news/DM-44670.feature.1.rst
+++ b/doc/news/DM-44670.feature.1.rst
@@ -1,0 +1,1 @@
+Update ATCalSys.prepare_for_flat to call setup_electrometer.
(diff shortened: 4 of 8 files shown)


python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -196,11 +196,38 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+            getattr(self.rem, component_name)
+            for component_name in self.components_attr
+            if "electrometer" in component_name
+        ]:
+            await electrometer.cmd_setMode.set_start(mode=mode)
+            await electrometer.cmd_setRange.set_start(setRange=range)
+            await electrometer.cmd_setIntegrationTime.set_start(
+                intTime=integration_time
+            )
+            await electrometer.cmd_performZeroCalib.start(timeout=self.long_timeout)
```
parfa30, line 223: Should this be `set_start` or just `start`? @tribeiro

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -226,19 +253,21 @@ async def run_calibration_sequence(
 
         calibration_type = getattr(CalibrationType, str(config_data["calib_type"]))
         if calibration_type == CalibrationType.WhiteLight:
-            calibration_wavelenghts = np.array([float(config_data["wavelength"])])
+            calibration_wavelengths = np.array([float(config_data["wavelength"])])
         else:
             wavelength = float(config_data["wavelength"])
             wavelength_width = float(config_data["wavelength_width"])
```
parfa30, line 259: Can we change `wavelength_width` to something like `filter_width` or `band_width`. I find this term confusing. And why not just have a `wavelength_start` and `wavelength_end` with resolution?

parfa30 (commented):
Thanks for putting in this function @elanaku. I think this should work. Just have a couple questions that maybe @tribeiro will have to answer.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
tribeiro: just start.. we only use `set_start` when we need to pass command parameters. Note that timeout is not a command parameter.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
tribeiro: I think this is a matter of personal preference. I find passing center wavelength + width more expressive. This is not a filter or band width, but rather the "scan width". I think we can certainly change these to something else if you think they will be more clear, but let's not do it as part of this ticket. Let's focus on what is being doing here instead. 

Feel free to open a ticket for updating this though, and we can discuss in the ticket what would be a better solution.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -196,11 +196,38 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+    async def setup_electrometers(
+        self, mode: int = 1, range: float = -1, integration_time: float = 0.1
+    ) -> None:
+        """Setup all electrometers."""
+        for electrometer in [
+            getattr(self.rem, component_name)
+            for component_name in self.components_attr
+            if "electrometer" in component_name
+        ]:
+            await electrometer.cmd_setMode.set_start(mode=mode)
```
tribeiro, line 218: Please, add a timeout to all the calls to `set_start` here and in the next 3 lines. You can use `timeout=self.long_timeout`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -196,11 +196,38 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+        )
+
         await asyncio.gather(
-            task_setup_monochromator,
-            task_setup_latiss,
+            task_setup_monochromator, task_setup_latiss, task_setup_electrometer
         )
 
+    async def setup_electrometers(
+        self, mode: int = 1, range: float = -1, integration_time: float = 0.1
```
tribeiro, line 210: do you really need the default values here? it seems like this is always getting values from `config_data` above. I suggest you remove the default values.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -196,11 +196,38 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+    ) -> None:
+        """Setup all electrometers."""
```
tribeiro, line 212: Please document the input parameters.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -196,11 +196,38 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+        task_setup_electrometer = self.setup_electrometers(
+            mode=int(config_data["electrometer_mode"]),
+            range=float(config_data["electrometer_range"]),
+            integration_time=float(config_data["electrometer_integration_time"]),
+        )
+
         await asyncio.gather(
-            task_setup_monochromator,
-            task_setup_latiss,
+            task_setup_monochromator, task_setup_latiss, task_setup_electrometer
```
tribeiro, line 206: This is a picky comment but do you mind adding a comma after `task_setup_electrometer`? This will cause `black` to format the code such that each task is in a separate line (like it was originally), thus reducing the signature change when we add new tasks.

tribeiro (commented):
Thanks for the updates Elana. I have a few comments I hope you will take into consideration before I can approve the PR. 

Also, can you please update the `test_setup_electrometers` in `tests/auxtel/test_atcalsys.py` to reflect the changes you've made? basically, make sure the commands you added are awaited as expected?

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -309,12 +318,10 @@ async def _take_data(
         exposures_done: asyncio.Future = asyncio.Future()
 
         fiber_spectrum_exposure_coroutine = self.take_fiber_spectrum(
-            exposure_time=fiber_spectrum_exposure_time,
-            exposures_done=exposures_done,
+            exposure_time=fiber_spectrum_exposure_time, exposures_done=exposures_done
```
tribeiro, line 321: please, undo this change here.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -309,12 +318,10 @@ async def _take_data(
...
         )
         electrometer_exposure_coroutine = self.take_electrometer_scan(
-            exposure_time=electrometer_exposure_time,
-            exposures_done=exposures_done,
+            exposure_time=electrometer_exposure_time, exposures_done=exposures_done
```
tribeiro, line 324: please, undo this change here.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -72,13 +72,37 @@ def __init__(
 
         self.calibration_config: dict[str, dict[str, typing.Any]] = dict()
 
-    async def setup_electrometers(self) -> None:
-        """Setup all electrometers."""
+    async def setup_electrometers(
+        self, mode: int, range: float, integration_time: float
```
tribeiro, line 76: if mode is an enumeration, please, use the enumeration instead of an int.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -72,13 +72,37 @@ def __init__(
...
+    ) -> None:
+        """Setup all electrometers.
+
+        Parameters
+        ----------
+        mode : `int`
```
tribeiro, line 82: if this is an enumeration, please use the enum instead

tribeiro (commented):
Here some additional comments. 

I think this is mostly ready to go. However, before I approve it, please, check my latest comments and cleanup the git history.

Also, one note on the news fragment. Since you are doing more than one change, the way to do that is by creating different files, one for each change. This will also allow you to properly classify them. It works like this:

> file doc/news/DM-44670.bugfix.rst

```
Update ATCalSys so that the filter scans have the correct wavelength range.
```
> file doc/news/DM-44670.feature.rst

```
Update BaseCalSys.setup_electrometers to setup electrometer mode, range, and integration time from input parameters.
```
> file doc/news/DM-44670.feature.1.rst

```
Update ATCalSys.prepare_for_flat to call setup_electrometer.
```

> file doc/news/DM-44670.feature.2.rst 

```
Update atcasys configuration such that the electrometer exposure time is similar to the camera exposure time and to include the additional electrometer configuration.
```

python/lsst/ts/observatory/control/base_calsys.py
elanaku: I think we do pass an int that is then turned into an enum by the controller in ts_electrometer (at least, that's my understanding - @parfa30 can confirm).  Should I make a note of that? Should I call this an enum value?

parfa30: yeah, I will add in the enumeration!

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -72,13 +73,40 @@ def __init__(
...
+        ----------
+        mode : `str`
+            Electrometer measurement mode.
+        range : `float`
+            Electrometer measurement range. -1 for autorange.
+        integration_time : `float`
+            Electrometer measurement range.
+        """
+        elec_mode = getattr(UnitToRead, mode)
+        elecMode = elec_mode.value
```
tribeiro, line 91: Please, merge these 2 lines into:

```
elec_mode = getattr(UnitToRead, mode).value
```

tribeiro (approved):
looks good! thanks for the updates!

PR #144, opened 2024-07-12
tickets/DM-45219

elanaku:
Added new configs to ATCalsys.yaml for the AT PTC curves.


python/lsst/ts/observatory/control/data/atcalsys.yaml
```diff
@@ -76,3 +76,251 @@ at_whitelight_r:
...
+  atspec_filter: SDSSr_65mm
+  atspec_grating: empty_1
+  wavelength: 500.0
+  monochromator_grating: RED
+  exit_slit: 5.0
+  entrance_slit: 5.0
+  electrometer_integration_time: 0.1
+  electrometer_mode: CURRENT
+  electrometer_range: -1
+  use_fiberspectrograph: true
```
tribeiro, line 91: These two flags are currently not used, but it is good to add them here as future proofing the configuration. 

A couple suggestions;

1. Can you please move these two to be close to `use_camera` above?
2. I think we want `use_fiberspectrograph: false`
3. Add the same fields to the other configurations.

tribeiro (approved):
This looks good. Please, see my inline comment before merging.

Also, please, add the news fragment.

PR #145, opened 2024-07-12
adding calculate optimized exposure times

parfa30:
This new function returns an exposure list with wavelength and camera, electrometer and fiberspectrograph exposure times. It is called in `run_calibration_sequence`. The main change to this function is that instead of looping through wavelength, it loops through exposure. This does change what is included in a "step", but I think for how we will be running sequences on AuxTel that is ok. 

There are changes to the configuration file `atcalsys.yaml`:
* Removal of exptime for electrometer and fiber spectrograph
* Addition of `use_fiberspectrograph` and `use_electrometer`

And finally, removed the use of `exposures_done` for both electrometer and fiber spectrograph. We will not want these exposures to just repeat. For one, if the electrometer exposure repeats it will significantly delay the sequence due to readout times and often won't give us useful information. Also, generally, we only want one fiber spectrograph exposure. If we want to continuously take them several times during an exposure, we should add in an explicit number.

### DIFF
--- a/python/lsst/ts/observatory/control/auxtel/atcalsys.py
+++ b/python/lsst/ts/observatory/control/auxtel/atcalsys.py
@@ -24,6 +24,7 @@ import asyncio
 import logging
 import time
 import typing
+from dataclasses import dataclass
 
 import numpy as np
 from lsst.ts import salobj, utils
@@ -67,6 +68,14 @@ class ATCalsysUsages(Usages):
                 self.DryTest,
             ]
         )
+
+
+@dataclass
+class ATCalsysExposure:
+    wavelength: float
+    camera: float
+    fiberspectrograph: float | None
+    electrometer: float | None
 
 
 class ATCalsys(BaseCalsys):
@@ -205,7 +214,7 @@ class ATCalsys(BaseCalsys):
 
     async def calculate_optimized_exposure_times(
         self, wavelengths: list, config_data: dict
-    ) -> list:
+    ) -> list[ATCalsysExposure]:
         """Calculates the exposure times for the fiber spectrograph and
         the electrometer given camera exposure times and other
         relevant configuration paramaters.
@@ -249,7 +258,7 @@ class ATCalsys(BaseCalsys):
 
             for i, exptime in enumerate(config_data["exposure_times"]):
                 exposures.append(
-                    (
+                    ATCalsysExposure(
                         wavelength,
                         exptime,
                         electrometer_exptimes[i],
@@ -257,23 +266,14 @@ class ATCalsys(BaseCalsys):
                     )
                 )
 
-        exposure_list = np.array(
-            exposures,
-            dtype=[
-                ("wavelength", float),
-                ("camera", float),
-                ("fiberspectrograph", float),
-                ("electrometer", float),
-            ],
-        )
-        return exposure_list
+        return exposures
 
     async def _calculate_electrometer_exposure_times(
         self,
         exptimes: list,
         electrometer_integration_time: float,
         use_electrometer: bool,
-    ) -> list:
+    ) -> list[float | None]:
         """Calculates the optimal exposure time for the electrometer
 
         Parameters
@@ -285,14 +285,14 @@ class ATCalsys(BaseCalsys):
 
         Returns
         -------
-        `list`
+        `list`[`float` | `None`]
             Exposure times for the electrometer
         """
         electrometer_buffer_size = 16667
         electrometer_integration_overhead = 0.00254
         electrometer_time_separation_vs_integration = 3.07
 
-        electrometer_exptimes = []
+        electrometer_exptimes: list[float | None] = []
         for exptime in exptimes:
             if use_electrometer:
                 time_sep = (
@@ -308,7 +308,7 @@ class ATCalsys(BaseCalsys):
                 else:
                     electrometer_exptimes.append(exptime)
             else:
-                electrometer_exptimes.append(np.nan)
+                electrometer_exptimes.append(None)
         return electrometer_exptimes
 
     async def _calculate_fiberspectrograph_exposure_times(
@@ -317,7 +317,7 @@ class ATCalsys(BaseCalsys):
         entrance_slit: float,
         exit_slit: float,
         use_fiberspectrograph: bool,
-    ) -> list:
+    ) -> list[float | None]:
         """Calculates the optimal exposure time for the electrometer
 
         Parameters
@@ -331,10 +331,10 @@ class ATCalsys(BaseCalsys):
 
         Returns
         -------
-        `list`
+        `list`[`float` | `None`]
             Exposure times for the fiberspectrograph
         """
-        fiberspectrograph_exptimes = []
+        fiberspectrograph_exptimes: list[float | None] = []
         for exptime in exptimes:
             if use_fiberspectrograph:
                 base_exptime = 1  # sec
@@ -347,7 +347,7 @@ class ATCalsys(BaseCalsys):
                 )
                 fiberspectrograph_exptimes.append(fiberspectrograph_exptime)
             else:
-                fiberspectrograph_exptimes.append(np.nan)
+                fiberspectrograph_exptimes.append(None)
         return fiberspectrograph_exptimes
 
     async def run_calibration_sequence(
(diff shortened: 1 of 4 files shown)


python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +213,81 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+                    base_exptime = 1  # sec
+                    entry_slit_multiplier = 6 / (config_data["entrance_slit"] + 1)
+                    exit_slit_multiplier = 6 / (config_data["exit_slit"] + 1)
+                    # TODO (DM-45235): Improve these multipliers with testing
+                    # Also, include some wavelength dependence based on testing
+                    fiberspectrograph_exptime = (
+                        base_exptime * entry_slit_multiplier * exit_slit_multiplier
+                    )
+                else:
+                    fiberspectrograph_exptime = np.nan
```
tribeiro, line 269: consider using `None` instead of `np.nan`

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +213,81 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+                    max_exp_time = 16667 * time_sep  # buffer size
+                    if exptime > max_exp_time:
+                        electrometer_exptime = max_exp_time
+                        self.log.info(
+                            f"Electrometer exposure time reduced to {max_exp_time}"
+                        )
+                    else:
+                        electrometer_exptime = exptime
+                else:
+                    electrometer_exptime = np.nan
```
tribeiro, line 257: Consider using `None` instead of `np.nan`, it has a smaller memory footprint and will make it easier to check it later.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +213,81 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+            List of exposure information, including wavelength and camera,
+            fiberspectrograph and electrometer exposure times.
+
+        """
+        exposures = []
+        for wavelength in wavelengths:
+            for exptime in config_data["exposure_times"]:
+                if config_data["use_electrometer"]:
+                    time_sep = (
+                        config_data["electrometer_integration_time"] * 3.07
```
tribeiro, line 246: Where are these hard coded numbers coming from? Do they have any meaning? I would suggest either making them local variables with names expressing what they are or making them class attributes. 

The same comment applies to the 3 numbers that appears here, e.g. `3.07`, `0.00254` and `16667`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +213,81 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+        -------
+        exposure_list : `array`
+            List of exposure information, including wavelength and camera,
+            fiberspectrograph and electrometer exposure times.
+
+        """
+        exposures = []
+        for wavelength in wavelengths:
+            for exptime in config_data["exposure_times"]:
+                if config_data["use_electrometer"]:
```
tribeiro, line 244: I suggest you extract the code inside these two `if` statements (`config_data["use_electrometer"]` and `if config_data["use_fiberspectrograph"])` into methods. Something like `_calculate_electrometer_exposure_times` and `_calculate_fiberspectrograph_exposure_times`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +213,81 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+        exposure_list = np.array(
+            [np.vstack(exposures)],
+            dtype=[
+                ("wavelength", float),
+                ("camera", float),
+                ("fiberspectrograph", float),
+                ("electrometer", float),
+            ],
+        )
+        return exposure_list
```
tribeiro, line 289: you are documenting the return type as a `list` but this is a `numpy` array no? Maybe this should be a list of dictionaries? Something like `list[dict[str, float]]`? then make each "exposure" be a dict with the elements you are now using the `dtype` for?

tribeiro (commented):
This is looking pretty good! I have some small inline comments for you to consider. Also, please, remember to add the news fragment.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -373,8 +448,7 @@ async def take_electrometer_scan(
 
         electrometer_exposures = list()
 
-        while not exposures_done.done():
-
+        if not math.isnan(exposure_time):
```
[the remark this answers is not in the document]
parfa30: Thought about using None, but went with np.nan because there is some code that make it float(exposure_time) and it's unhappy. I'll get rid of `math` though!

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: As mentioned above, the reason for using np.nan instead of None is that there are several places that have float(exposure_time) and np.nan will not fault. If it takes up too much memory I can refactor things, but this seemed a good way to keep most of the other code intact.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
parfa30: I went back and forth on the best kind of data type to use here and finally landed on a numpy rec array. To me, it's a bit cleaner than a list of dictionaries.

tribeiro: The numpy array is really useful for data analysis type applications, but for this one, not so much. If you think about the two alternatives, you are basically generating a list of wavelength, camera, fiber spec, electrometer exposure times, so a list of something. Whereas the numpy array is a table with different columns. I think the list of dictionary makes a lot more sense. 

Also, note, for instance, you can't provide a good type hint for  the data when it is a numpy array. With a list of dict, it is more clear. 

Also, if you want to take this one step further, you could make a data class to store this information and make the return type a list of this object instead. That would be even better. It would look like this:

```
@dataclass
class ATCalsysExposure:
    wavelength: float
    camera: float
    fiberspectrograph: float | None
    electrometer: float | None
```

and the return type would be `list[ ATCalsysExposure ]`.

Note that here I am also assuming you will use `None` for when there's no exposure of a particular type.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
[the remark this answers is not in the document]
tribeiro: You can make the code that uses `float(exposure_time)` happy with something like:

```
float(exposure_time) if exposure_time is not None else None
```

This will probably also require changing some type annotations to make sure whatever is receiving the float, also accepts a `None`, e.g.:

```
    async def _take_data(
        ...
        fiber_spectrum_exposure_time: float | None,
        electrometer_exposure_time: float | None,
    ) -> dict:
```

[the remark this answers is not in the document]
parfa30: Do you feel strongly about not using np.nan? I can make these changes but they aren't necessary if we use np.nan

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -69,6 +70,14 @@ def __iter__(self) -> typing.Iterator[int]:
         )
 
 
+@dataclass
+class ATCalsysExposure:
```
tribeiro, line 74: Please, add a docstring to this class explaining what it is for and also document the attributes, e.g.;

```
@dataclass
class ATCalsysExposure:
    """Store Exposure information for ATCalsys.

    Attributes
    ----------
    wavelength : float
        Wavelength of the calibration exposure, in nm.
    camera : float
        Camera exposure time, in sec.
    fiberspectrograph : float | None
        Fiber spectrograph exposure time, in sec.
        If None, skip fiber spectrograph acquisition.
    electrometer : float | None
        Electrometer exposure time, in sec.
        If None, skip electrometer acquisition.
    """
    ...
 ```

File: python/lsst/ts/observatory/control/auxtel/atcalsys.py [hunk 8]
```diff
@@ -208,6 +212,144 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+        Parameters
+        ----------
+        wavelengths : `list`
+            List of all wavelengths for this exposure list
+        config_data : `dict`
+            All information from configuration file
+
+        Returns
+        -------
+        exposure_list : `array`
```
tribeiro, line 235: return type is now `list[ATCalsysExposure]`.

python/lsst/ts/observatory/control/auxtel/atcalsys.py
```diff
@@ -208,6 +212,144 @@ async def prepare_for_flat(self, config_name: str) -> None:
...
+            List of exposure information, including wavelength and camera,
+            fiberspectrograph and electrometer exposure times.
+
+        """
+
+        exposures = []
```
tribeiro, line 241: You might want to provide the type annotation here. Mypy is probably able to derive it from context, but it doesn't hurt to help it out. It also let whoever is reading the code know that this variable has the same type as the return type, so it is probably the variable you are returning.

tribeiro (approved):
This looks great! I am leaving 3 really minor inline comments I hope you will consider before merging.

PR #147, opened 2024-07-24
DM-39830: Implement support for initGuiders command in OCS.

tribeiro:


python/lsst/ts/observatory/control/utils/roi_spec.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class ROICommon(BaseModel):
+    rows: int = Field(default=50)
+    cols: int = Field(default=50)
+    integrationTimeMillis: int = Field(alias="integration_time_millis", default=100)
+
+    @validator("rows")
+    def check_rows(cls, v: int) -> int:
+        if v < 0:
```
edennihy, line 40: Is there any scenario where 0 is a valid minimum value? If not, maybe we should make these checks if v < 1.

python/lsst/ts/observatory/control/utils/roi_spec.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError(f"Minimum rows must be positive, not {v!r}.")
+        return v
+
+    @validator("cols")
+    def check_cols(cls, v: int) -> int:
+        if v < 0:
```
edennihy, line 46: Same as above.

python/lsst/ts/observatory/control/utils/roi_spec.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError(f"Minimum cols must be positive, not {v!r}.")
+        return v
+
+    @validator("integrationTimeMillis")
+    def check_integrationTimeMillis(cls, v: int) -> int:
+        if v < 0:
```
edennihy, line 52: Same as above. I think 0 here would correspond to bias..

edennihy (approved):
Minor comments to consider. Otherwise, approved.

doc/news/DM-39830.feature.rst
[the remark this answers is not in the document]
tribeiro: uhm.. I am not sure, I think this is all in the category of "support init guiders".

PR #155, opened 2024-08-26
DM-45975: update grating for whitelight in PTC curves

parfa30:
We updated the Monochromator xml enumeration (DM-45474) so that it correctly aligns with the gratings. I have updated the configuration for the PTC curves so that it calls on MIRROR instead of the BLUE grating.


tribeiro (approved):
looks good.

PR #158, opened 2024-08-30
Add configuration schema validation support to BaseCalsys and schema files for ATCalsys and MTCalsys classes

MarcoRocchietti:


python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -26,6 +26,7 @@
 import logging
 import typing
 
+import jsonschema
```
tribeiro, line 29: Instead of using `jsonschema` directly, I would recommend you use `salobj.DefaultingValidator` as it allows you to add default values for the parameters, reducing the amount of duplicated needed in the configuration.

This is what we use for [SAL Scripts](https://github.com/lsst-ts/ts_salobj/blob/4841de22e72e6e8bc95442d289e9afafa7bb248e/python/lsst/ts/salobj/base_script.py#L116).

You can check the link above to see how we use it.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -188,6 +205,25 @@ def load_calibration_config_file(self, filename: str | None = None) -> None:
         with open(data_path, "r") as f:
             self.calibration_config = yaml.safe_load(f)
 
+        config_schema = None
+        try:
```
tribeiro, line 209: we are providing the schema ourselves so there's no reason to do this try/except here. If we can't open the file it should error out.

python/lsst/ts/observatory/control/base_calsys.py
```diff
@@ -188,6 +205,25 @@ def load_calibration_config_file(self, filename: str | None = None) -> None:
...
+                config_schema = yaml.safe_load(f)
+        except FileNotFoundError:
+            self.log.warning(
+                f"Configuration schema file {schema_path} not found. "
+                f"No schema validation is performed for {data_path}."
+            )
+
+        if config_schema:
+            for item in self.calibration_config:
+                try:
```
tribeiro, line 220: I don't think re-raising the exception like this is adding anything useful. What you could do instead is try to validate all sessions and then raise an exception at the end aggregating all the failures.

Something like this:

```
validation_errors = ""
for item in self.calibration_config:
    try:
        # validate config
        ...
    except jsonschema.ValidationError as e:
        validation_errors += f"\t{item} failed validation: {e.message}.\n"
        self.log.exception(f"{item} failed validation.")
if validation_errors:
    raise RuntimeError(f"Failed schema validation:\n{validation_errors}Check logs for more information.")
```

tribeiro (approved):
Looks great! thanks!

python/lsst/ts/observatory/control/data/atcalsys_schema.yaml
```diff
@@ -0,0 +1,55 @@
...
+  atspec_grating:
+    type: string
+  wavelength:
+    type: number
+  wavelength_width:
+    type: number
+  wavelength_resolution:
+    type: number
+  monochromator_grating:
+    type: string
```
tribeiro, line 21: I think this one is also an enumeration.

python/lsst/ts/observatory/control/data/atcalsys_schema.yaml
```diff
@@ -0,0 +1,55 @@
...
+  exit_slit:
+    type: number
+  entrance_slit:
+    type: number
+  electrometer_integration_time:
+    type: number
+  electrometer_mode:
+    type: string
```
tribeiro, line 29: enumeration as well?

python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
```diff
@@ -0,0 +1,56 @@
+$schema: http://json-schema.org/draft-07/schema#
+$id: https://github.com/lsst-ts/ts_observatory_control/blob/develop/python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
+title: Configuration file validation schema of Main Telescope Calibration System
+type: object
+properties:
+  calib_type:
+    type: string
```
tribeiro, line 7: This should be an enumeration.

python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
```diff
@@ -0,0 +1,56 @@
...
+  use_camera:
+    type: boolean
+  mtcamera_filter:
+    type: string
+  led_name:
+    type: array
+    items:
+      type: string
```
tribeiro, line 15: I think this is also an enumeration.

python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
```diff
@@ -0,0 +1,56 @@
...
+  use_electrometer:
+    type: boolean
+  use_fiberspectrograph_red:
+    type: boolean
+  use_fiberspectrograph_blue:
+    type: boolean
+  electrometer_integration_time:
+    type: number
+  electrometer_mode:
+    type: string
```
tribeiro, line 35: enumeration?

tribeiro (changes requested):
I added some comments about string fields that are actually enumerations.

Also, can you review the current configuration and see if there are defaults you could use for any of the parameters? For example, if there is a parameter that has the same value in all configuration, that value can probably be converted into a default value and specified in the schema instead of letting it be repeated everywhere. Also, if there are values that are used is most cases and only change in a couple configuration, they are good candidates to have default values. 

Default values is something that our `DefaultingValidator` provides that is not part of the jsonschema standard, but it works really well. All you need to do is add a `default: value` field in the schema, e.g.;

```
  use_camera:
    type: boolean
    default: true
```

Also, please, consider adding a "description" field to all the items in the schema. Let me know if you can't figure out what the description of a parameter should be from context and I can help you out.

MarcoRocchietti:
> I added some comments about string fields that are actually enumerations.

I found a couple more attributes that can be defined as enumerations (marked with \*):

- For `atcalsys.yaml`:
  - calib_type
  - atspec_filter (\*)
  - atspec_grating (\*)
  - monochromator_grating
  - electrometer_mode
- For `mtcalsys.yaml`:
  - calib_type
  - mtcamera_filter (*)
  - led_name
  - electrometer_mode
  - laser_mode (*)

I have updated [atcalsys_schema.yaml](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/data/atcalsys_schema.yaml) and [mtcalsys_schema.yaml](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml). Let me know if I have missed any values. I have done an extensive search and believe I’ve covered most of them. The attribute `led_name` only includes the values I’ve observed in the configurations, as I couldn’t find an enumeration definition or technical specification for the LED names.

Regarding `laser_mode`, it’s not currently present in any configuration, but it’s expected to be loaded in the [MTCalsys.setup_calsys](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/maintel/mtcalsys.py#L243) method. A problem I noticed is that the [MTCalsys.setup_laser](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/maintel/mtcalsys.py#L263) method expects an integer value instead of a string for the constants: `NONPROPAGATING_CONTINUOUS_MODE`, `NONPROPAGATING_BURST_MODE`, `PROPAGATING_CONTINUOUS_MODE`, and `PROPAGATING_BURST_MODE`. Also, the [method's docstring](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/maintel/mtcalsys.py#L258) indicates that the parameter `mode` should be one of CONTINUOUS, BURST, or TRIGGER, which doesn’t align well with the actual enum `LaserDetailedState`. I’m not sure how to specify this attribute in the validation schema. If I define it as a string enum (using the actual enumeration names of `LaserDetailedState`), the method will fail [when reading the value](https://github.com/lsst-ts/ts_observatory_control/blob/1151b8a9525e6604103a2f966e9791fc2f035a31/python/lsst/ts/observatory/control/maintel/mtcalsys.py#L277). But I also don’t think it makes sense for the configuration file to include integer values there. It will be necessary to adjust how the `MTCalsys.setup_laser` method handles this attribute if the validation schema defines `laser_mode` as an enumeration. Because of that issues for the moment i'm not including the attribute in the `mtcalsys_schema.yaml` validation schema.

MarcoRocchietti:
> Also, can you review the current configuration and see if there are defaults you could use for any of the parameters? For example, if there is a parameter that has the same value in all configuration, that value can probably be converted into a default value and specified in the schema instead of letting it be repeated everywhere. Also, if there are values that are used is most cases and only change in a couple configuration, they are good candidates to have default values

Ok, so for those attributes where I define default values in the validation schema, the idea is to update the `atcalsys.yaml` and `mtcalsys.yaml` files to only specify the attribute when it is not the default value, right?

Ok, so for those attributes where I define default values in the validation schema, the idea is to update the atcalsys.yaml and mtcalsys.yaml files to only specify the attribute when it is not the default value, right?

python/lsst/ts/observatory/control/data/atcalsys_schema.yaml
```diff
@@ -0,0 +1,77 @@
...
+  calib_type:
+    type: string
+    enum:
+      - Mono
+      - WhiteLight
+  use_camera:
+    type: boolean
+  atspec_filter:
+    type: string
+    enum:
```
tribeiro, line 15: These here are not enums. They can change to other things beyond these values, so you can leave it as a free form string.

python/lsst/ts/observatory/control/data/atcalsys_schema.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - SDSSr_65mm
+      - SDSSg_65mm
+      - SDSSz_65mm
+      - SDSSy_65mm
+  atspec_grating:
+    type: string
+    enum:
```
tribeiro, line 22: Same as filter, this is not an enum.

python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
```diff
@@ -0,0 +1,82 @@
...
+  calib_type:
+    type: string
+    enum:
+      - Mono
+      - WhiteLight
+  use_camera:
+    type: boolean
+  mtcamera_filter:
+    type: string
+    enum:
```
tribeiro, line 15: This is not an enum

tribeiro (approved):
Alright, looks good. I will approve the PR, but please, check my comments, filter and grating are not enumerations.

PR #159, opened 2024-09-05
tickets/DM-45609

iglesu:
- Add the `park_dome` method in `MTCS` to park the dome. This includes:
  - Ensuring the dome is enabled before initiating the park sequence.
  - Commanding the dome to park and waiting for the dome's azimuth motion state to reach `PARKED`.
  - Added a polling mechanism to periodically check the dome's `azMotion` event until it reaches the `PARKED` state or a timeout occurs.
  - Raised an error if the dome fails to park within the timeout window.

- Updated the `MTCSAsyncMock` class to simulate dome parking behavior:
  - Mocked the `evt_azMotion` event to simulate the dome reaching the `PARKED` state with `inPosition=True`.
  - Added mocking for the `cmd_park` command to simulate a dome park action.

- Added a new unit test `test_park_dome`
  - Asserted that the `evt_azMotion` event is awaited and reaches the `PARKED` state with `inPosition=True`.

### DIFF
--- a/doc/news/DM-45609.feature.rst
+++ b/doc/news/DM-45609.feature.rst
@@ -1,0 +1,1 @@
+Implement dome parking in MTCS.
--- a/python/lsst/ts/observatory/control/maintel/mtcs.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcs.py
@@ -31,8 +31,8 @@ import astropy.units as u
 import numpy as np
 from astropy.coordinates import Angle
 from lsst.ts import salobj, utils
-from lsst.ts.idl.enums import MTM1M3, MTM2, MTPtg, MTRotator
 from lsst.ts.utils import angle_diff
+from lsst.ts.xml.enums import MTM1M3, MTM2, MTDome, MTPtg, MTRotator
 
 try:
     from lsst.ts.xml.tables.m1m3 import FATable
@@ -140,6 +140,10 @@ class MTCS(BaseTCS):
         self.dome_flat_az = 20.0
         self.dome_flat_el = self.dome_park_el
         self.dome_slew_tolerance = Angle(1.5 * u.deg)
+
+        # TODO (DM-45609): This is an initial guess for the time it takes the
+        #  dome to park. It might need updating.
+        self.park_dome_timeout = 600
 
         self._dome_az_in_position: typing.Union[None, asyncio.Event] = None
         self._dome_el_in_positio: typing.Union[None, asyncio.Event] = None
@@ -508,9 +512,9 @@ class MTCS(BaseTCS):
         Parameters
         ----------
         timeout: `float`
-            How to to wait for mount to be in position (in seconds).
+            How to wait for mount to be in position (in seconds).
         wait_settle: `bool`
-            After receiving the in position command add an addional settle
+            After receiving the in position command, add an additional settle
             wait? (default: True)
 
         Returns
@@ -578,6 +582,91 @@ class MTCS(BaseTCS):
         """
         await self._dome_el_in_position.wait()
         return "Dome elevation in position."
+
+    async def wait_for_dome_state(
+        self,
+        expected_states: set[MTDome.MotionState],
+        bad_states: set[MTDome.MotionState],
+        timeout: float,
+        check_in_position: bool = False,
+    ) -> None:
+        """Wait for a specific dome state.
+
+        Parameters
+        ----------
+        expected_states : set[MTDome.MotionState]
+            Valid states to transition into while un-parking.
+        bad_states : set[MTDome.MotionState]
+            States that are not allowed while un-parking and should raise an
+            error.
+        timeout : float
+            Maximum time to wait for the correct state.
+
+        Raises
+        ------
+        RuntimeError
+            If a bad state is encountered or the expected state is not reached
+             in time.
+        """
+
+        def dome_ready(az_motion: salobj.type_hints.BaseMsgType) -> bool:
+            return (
+                az_motion.state in expected_states and az_motion.inPosition
+                if check_in_position
+                else az_motion.state in expected_states
+            )
+
+        az_motion = await self.rem.mtdome.evt_azMotion.aget(timeout=timeout)
+
+        while not dome_ready(az_motion):
+            az_motion = await self.rem.mtdome.evt_azMotion.next(
+                Flush=False, timeout=timeout
+            )
+
+            if az_motion.state in bad_states:
+                raise RuntimeError(
+                    f"Dome transitioned to an invalid state: {MTDome.MotionState(az_motion.state).name}"
+                )
+
+            self.log.debug(
+                f"Dome state: {MTDome.MotionState(az_motion.state).name}, inPosition: {az_motion.inPosition}"
+            )
+
+    async def park_dome(self) -> None:
+        """Park the dome by moving it to the park azimuth."""
+        self.log.info("Parking dome")
+
+        await self.assert_all_enabled(
+            message="All components need to be enabled for parking the Dome."
+        )
+
+        # check first if Dome is already in PARKED state
+        az_motion = await self.rem.mtdome.evt_azMotion.aget(timeout=self.fast_timeout)
+
+        if az_motion.state == MTDome.MotionState.PARKED:
+            self.log.info("Dome is already in PARKED state.")
+        else:
+            self.rem.mtdome.evt_azMotion.flush()
+
+            await self.rem.mtdome.cmd_park.start(timeout=self.long_timeout)
+
+            # Define expected and bad states for parking
+            expected_states = {MTDome.MotionState.PARKED}
+            bad_states = {
+                MTDome.MotionState.ERROR,
+                MTDome.MotionState.UNDETERMINED,
+                MTDome.MotionState.DISABLED,
+                MTDome.MotionState.DISABLING,
+            }
+
+            self.log.info("Waiting for dome to reach the PARKED state.")
+
+            await self.wait_for_dome_state(
+                expected_states,
+                bad_states,
+                timeout=self.park_dome_timeout,
+                check_in_position=True,
+            )
 
     def set_azel_slew_checks(self, wait_dome: bool) -> typing.Any:
         """Handle azEl slew to wait or not for the dome.
--- a/python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
+++ b/python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
@@ -29,6 +29,7 @@ from lsst.ts import idl, utils
 from lsst.ts.idl.enums import MTM1M3
 from lsst.ts.observatory.control.maintel.mtcs import MTCS, MTCSUsages
 from lsst.ts.observatory.control.mock import RemoteGroupAsyncMock
+from lsst.ts.xml.enums import MTDome
 
 
 class MTCSAsyncMock(RemoteGroupAsyncMock):
@@ -117,6 +118,11 @@ class MTCSAsyncMock(RemoteGroupAsyncMock):
         self._mtdome_tel_light_wind_screen = types.SimpleNamespace(
             positionActual=0.0,
             positionCommanded=0.0,
+        )
+
+        # MTDome Motion PARKED state. State is set by the mocked park cmd.
+        self._mtdome_evt_azMotion_state = types.SimpleNamespace(
+            state=MTDome.MotionState.UNDETERMINED, inPosition=False
         )
 
         # MTM1M3 data
@@ -229,6 +235,9 @@ class MTCSAsyncMock(RemoteGroupAsyncMock):
         mtdome_mocks = {
             "tel_azimuth.next.side_effect": self.mtdome_tel_azimuth_next,
             "tel_lightWindScreen.next.side_effect": self.mtdome_tel_light_wind_screen_next,
+            "cmd_park.start.side_effect": self.mtdome_cmd_park,
+            "evt_azMotion.aget.side_effect": self.mtdome_evt_az_motion_state_next,
+            "evt_azMotion.next.side_effect": self.mtdome_evt_az_motion_state_next,
         }
 
         self.mtcs.rem.mtdome.configure_mock(**mtdome_mocks)
@@ -460,6 +469,23 @@ class MTCSAsyncMock(RemoteGroupAsyncMock):
         self, *args: typing.Any, **kwargs: typing.Any
     ) -> types.SimpleNamespace:
         return self._mtdome_tel_light_wind_screen
+
+    async def mtdome_cmd_park(self, timeout: float) -> None:
+        asyncio.create_task(self._mtdome_park())
+
+    async def _mtdome_park(self) -> None:
+        # Mock implementation of cmd_park
+        await asyncio.sleep(self.heartbeat_time)
+        self.log.info("Dome park command executed")
+        self._mtdome_evt_azMotion_state = types.SimpleNamespace(
+            state=MTDome.MotionState.PARKED, inPosition=True
+        )
+
+    async def mtdome_evt_az_motion_state_next(
+        self, *args: typing.Any, **kwargs: typing.Any
+    ) -> types.SimpleNamespace:
+        await asyncio.sleep(self.heartbeat_time * 3)
+        return self._mtdome_evt_azMotion_state
 
     async def mtm1m3_evt_detailed_state(
         self, *args: typing.Any, **kwargs: typing.Any
--- a/tests/maintel/test_mtcs.py
+++ b/tests/maintel/test_mtcs.py
@@ -32,6 +32,7 @@ from lsst.ts import idl, utils
 from lsst.ts.idl.enums import MTM1M3, MTM2
 from lsst.ts.observatory.control.mock.mtcs_async_mock import MTCSAsyncMock
 from lsst.ts.observatory.control.utils import RotType
+from lsst.ts.xml.enums import MTDome
 
 
 class TestMTCS(MTCSAsyncMock):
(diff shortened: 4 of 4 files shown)


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,37 @@ async def dome_el_in_position(self) -> str:
         await self._dome_el_in_position.wait()
         return "Dome elevation in position."
 
+    async def wait_for_dome_ready(self) -> None:
```
tribeiro, line 582: This method is only checking that the dome is in enables state. 

We usually do that at a different layer, using the [assert_all_enabled](https://github.com/lsst-ts/ts_observatory_control/blob/9e9ab43b4474a69c4e12723776bfbb1c55e9a023/python/lsst/ts/observatory/control/remote_group.py#L759) method. 

I think what you want to do here is check that the dome is not parked already, and that it is in a state that can be parked. So probably check the azEnabled or azMotion events.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,37 @@ async def dome_el_in_position(self) -> str:
...
+
+    async def park_dome(self) -> None:
+        """Park the dome by moving it to the park azimuth."""
+        self.log.info(f"Parking dome at azimuth {self.dome_park_az}.")
+
+        # Ensure the dome is ready for a move
+        await self.wait_for_dome_ready()
+
+        # Move the dome to the park position
+        await self.rem.mtdome.cmd_park.set_start(timeout=self.long_timeout)
```
tribeiro, line 600: Use `start` instead of `set_start` if you are not setting any attribute besides timeout.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,37 @@ async def dome_el_in_position(self) -> str:
...
+
+        # Wait for the dome to emit the evt_azMotion event in the PARKED state
+        self.log.info("Waiting for dome to reach the PARKED status.")
+
+        az_motion = await self.rem.mtdome.evt_azMotion.next()
```
tribeiro, line 605: you will need to make this check in a loop because there may be intermediary states between the current state and PARKED. You will need to define "valid states" (which are ok to transition to while parking) and "bad states" (which are not ok to transition to while parking and should raise an exception).  

See, for instance, how this is done in [_wait_for_mtm1m3_detailed_state](https://github.com/lsst-ts/ts_observatory_control/blob/9e9ab43b4474a69c4e12723776bfbb1c55e9a023/python/lsst/ts/observatory/control/maintel/mtcs.py#L871).

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -119,6 +120,11 @@ async def setup_types(self) -> None:
             positionCommanded=0.0,
         )
 
+        # MTDome Motion PARKED state. State is set by the mocked park cmd.
+        self._mtdome_evt_azMotion_state = types.SimpleNamespace(
+            state=None, inPosition=False
```
tribeiro, line 125: I don't think state can ever be `None`. It might be 0, but most likely it is going to be an instance of [MotionState](https://github.com/lsst-ts/ts_xml/blob/a16574f2a257c25557144337a94d9ba50605b017/python/lsst/ts/xml/enums/MTDome.py#L82).

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -461,6 +470,33 @@ async def mtdome_tel_light_wind_screen_next(
     ) -> types.SimpleNamespace:
         return self._mtdome_tel_light_wind_screen
 
+    async def mtdome_cmd_park(self, *args: typing.Any, **kwargs: typing.Any) -> None:
+        asyncio.create_task(self._mtdome_park())
+
+    async def _mtdome_park(self, *args: typing.Any, **kwargs: typing.Any) -> None:
```
tribeiro, line 476: you are not passing anything to this method so you can probably remove the arguments.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -461,6 +470,33 @@ async def mtdome_tel_light_wind_screen_next(
...
+        await asyncio.sleep(self.heartbeat_time)
+        self.log.info("Dome park command executed")
+        self._mtdome_evt_azMotion_state = types.SimpleNamespace(
+            state=MTDome.MotionState.PARKED, inPosition=True
+        )
+
+    async def mtdome_evt_az_motion_state_next(
+        self, *args: typing.Any, **kwargs: typing.Any
+    ) -> types.SimpleNamespace:
+        return await asyncio.create_task(
```
tribeiro, line 487: you don't need to create a task here, just do:

```
await asyncio.sleep(self.heartbeat_time * 3)
return self._mtdome_evt_azMotion_state
```

and remove the `_mtdome_evt_az_motion_state_next` method.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -229,6 +235,9 @@ async def setup_mtdome(self) -> None:
         mtdome_mocks = {
             "tel_azimuth.next.side_effect": self.mtdome_tel_azimuth_next,
             "tel_lightWindScreen.next.side_effect": self.mtdome_tel_light_wind_screen_next,
+            "cmd_park.set_start.side_effect": self.mtdome_cmd_park,
```
tribeiro, line 238: this should probably be just `start` instead of `set_start`, see my comment in the code.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,37 @@ async def dome_el_in_position(self) -> str:
...
+
+        # Check if the dome is in the PARKED state and in position
+        if az_motion.state == MTDome.MotionState.PARKED and az_motion.inPosition:
+            self.log.info("Dome parked successfully.")
+        else:
+            raise RuntimeError("Dome did not reach the park azimuth.")
```
tribeiro, line 611: When failing, make sure you report the dome state, e.g.:

```
raise RuntimeError(f"Unexpected dome azimuth motion state {MTDome.MotionState(az_motion.state).name}. Failed to reach the park position.")
```

tribeiro (changes requested):
Here some comments for you to consider.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: I precisely treat None as a non-existing state. The point was that this event state is set a some point during a task.

But I will rethink it again.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: Creating a task was the only way to avoid a race condition, if I don't do a task, the event seems to set to the original state. We need the state to change from one command, say park to another say unpark and the only way I managed that was via a task.

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: 7390752ef6738cd30f1ee84c9e65ffc1a28a7aaf

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: 7390752ef6738cd30f1ee84c9e65ffc1a28a7aaf

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: 592b7c8c01531a548501928ab2d23e4932d44244

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: bba97f5aae2eb4f18a5daf91961d9b935213a306

doc/news/DM-45609.feature.rst
```diff
@@ -0,0 +1 @@
+Implement dome parking.
```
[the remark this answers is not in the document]
iglesu: 5849716149f5f1ecda16d50c1fd0fd0cd57e50b7

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: bba97f5aae2eb4f18a5daf91961d9b935213a306

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: 07a452e0bd700b6b517dbb7c47bdc3e4f90c19be

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: 34b8c780a51499a3402d50cd51c01ff6b71c7ded

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: I considered the azEnabled but looking at the mtdome_csc.py it looks to me they use it more for internal checks.  In the do_park command they check for the summary state to be Enabled. 

The other thing to consider is that in the mtcs park_dome method I check the dome is not parked already but In that  test I don't check for the inPosition state to be True. I do that later when we command the dome to park, then we check for the PARKED state and also to be in position.

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: f358ad671ca65d197dd14207905df0efe2b4930c

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+
+            # If check_in_position is True, verify the dome is in the correct
+            # position.
+            if check_in_position and not az_motion.inPosition:
+                self.log.debug(
+                    f"Dome not yet in position. Current state: " f"{az_motion.state}."
+                )
+
+            self.log.debug(
+                f"Dome state: {az_motion.state}, inPosition: " f"{az_motion.inPosition}"
```
tribeiro, line 626: See my comment about converting the `az_motion.state` into a `MTDome.MotionState` enumeration to print its name.

Also, looks like black is putting the 2 strings in the same line, so you probably need only 1 string.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            az_motion = await self.rem.mtdome.evt_azMotion.next(
+                Flush=False, timeout=timeout
+            )
+
+            if az_motion.state in bad_states:
+                raise RuntimeError(
+                    f"Dome transitioned to an invalid state:" f" {az_motion.state}"
+                )
+
+            # If check_in_position is True, verify the dome is in the correct
```
tribeiro, line 618: See my comment below about integrating this check into the `while` condition.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            # expected state.
+        if check_in_position and not az_motion.inPosition:
+            raise RuntimeError(
+                f"Dome reached the expected state "
+                f"{az_motion.state} but is not yet in position."
+            )
+
+    async def park_dome(self) -> None:
+        """Park the dome by moving it to the park azimuth."""
+        self.log.info(f"Parking dome at azimuth {self.dome_park_az}.")
```
tribeiro, line 639: I think the MTDome has its own internal dome park position which could be different from `dome_park_az`. Consider making this only "Parking dome."

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            message="All components need to be enabled for parking the Dome."
+        )
+
+        # check first if Dome is already in PARKED state
+        az_motion = await self.rem.mtdome.evt_azMotion.aget(timeout=self.fast_timeout)
+
+        if az_motion.state == MTDome.MotionState.PARKED:
+            self.log.info("Dome is already in PARKED state.")
+        else:
+            # Move the dome to the park position
```
tribeiro, line 652: This command does more than just move the dome to the park position, please, consider removing this.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            # Define expected and bad states for parking
+            expected_states = {MTDome.MotionState.PARKED}
+            bad_states = {
+                MTDome.MotionState.ERROR,
+                MTDome.MotionState.UNDETERMINED,
+                MTDome.MotionState.DISABLED,
+                MTDome.MotionState.DISABLING,
+            }
+
+            # Wait for the dome to emit the evt_azMotion event in the PARKED
```
tribeiro, line 664: The log message is already giving this information. Please, consider removing the redundant comment.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            # state
+            self.log.info("Waiting for dome to reach the PARKED state.")
+
+            # Wait for the dome to park, ensuring it doesn't reach an invalid
```
tribeiro, line 668: I think it is clear from the method name that this is what you are doing already. Consider removing this comment.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            # state
+            await self.wait_for_dome_state(
+                expected_states,
+                bad_states,
+                timeout=self.long_timeout,
```
tribeiro, line 673: I think the `long_timeout` might not be long enough for this operations. I think this might require its own timeout, something like `dome_park_timeout`? And I think this needs to be something like 5-10 minutes. I am not sure if this is enough to park the dome but I think it is a good start. Maybe add a comment when you create this attribute that this might need updating in the future.

tests/maintel/test_mtcs.py
```diff
@@ -764,6 +765,21 @@ async def test_offset_xy_absorb(self) -> None:
...
         )
 
+    async def test_park_dome(self) -> None:
+        await self.mtcs.enable()
+        await self.mtcs.assert_all_enabled()
+
+        # Call the park_dome method
+        await self.mtcs.park_dome()
+
+        az_motion = await self.mtcs.rem.mtdome.evt_azMotion.aget()
```
tribeiro, line 775: I think here you want to check that the commands where called with the parameters you expect and also the events where awaited in the way you expect.

Something like

```
self.mtcs.rem.mtdome.cmd_park.start.assert_awaited_with(timeout=self.mtcs.park_dome_timeout)
```

And something similar for the events you await.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+        )
+
+        # check first if Dome is already in PARKED state
+        az_motion = await self.rem.mtdome.evt_azMotion.aget(timeout=self.fast_timeout)
+
+        if az_motion.state == MTDome.MotionState.PARKED:
+            self.log.info("Dome is already in PARKED state.")
+        else:
+            # Move the dome to the park position
+            await self.rem.mtdome.cmd_park.start(timeout=self.long_timeout)
```
tribeiro, line 653: Here is a "life hack" when handling commands that you will later inspect events for.

Make sure you call "flush" in the events you are going to inspect before you send the command. This will ensure the sample queue is cleared before you run the command and your later inspection won't be affected by old data. In this case, it should be something like this:

```
self.rem.mtdome.evt_azMotion.flush()
```

just before you issue the command.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+            raise RuntimeError(
+                f"Dome reached the expected state "
+                f"{az_motion.state} but is not yet in position."
+            )
+
+    async def park_dome(self) -> None:
+        """Park the dome by moving it to the park azimuth."""
+        self.log.info(f"Parking dome at azimuth {self.dome_park_az}.")
+
+        # Ensure the dome is ready for a move
```
tribeiro, line 641: I think you can remove this command here as well.

tribeiro (commented):
Here some additional comments.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +579,101 @@ async def dome_el_in_position(self) -> str:
...
+                    f"Dome not yet in position. Current state: " f"{az_motion.state}."
+                )
+
+            self.log.debug(
+                f"Dome state: {az_motion.state}, inPosition: " f"{az_motion.inPosition}"
+            )
+
+            # If we reach this point, the dome has successfully reached an
+            # expected state.
+        if check_in_position and not az_motion.inPosition:
```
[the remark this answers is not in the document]
iglesu: In my view check_in_position only makes sense if the state is valid, hence doing that after the while loop, but keeping in line with your suggestion this is my version of the while loop:

```
 while az_motion.state not in expected_states or (
                check_in_position and az_motion.state in expected_states and not az_motion.inPosition):
```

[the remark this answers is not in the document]
tribeiro: not sure I understand you comment. If the command is not in `expected_states` python will not even consider the second part of the `and` block so I think my original suggestion would take your concern into account.

python/lsst/ts/observatory/control/maintel/mtcs.py
tribeiro: This comment is still here.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -579,6 +583,92 @@ async def dome_el_in_position(self) -> str:
...
+            Maximum time to wait for the correct state.
+
+        Raises
+        ------
+        RuntimeError
+            If a bad state is encountered or the expected state is not reached
+             in time.
+        """
+
+        # Helper function checks if dome's ready.
```
tribeiro, line 612: I am not sure how useful this comment is. I think the method name is descriptive enough that it is clear what this method is for. If you really want to make a comment, make it a docstring.

iglesu: Removed, and sorry for the extra comments, I'll be more sparse going forward.

tribeiro (approved):
looks good.

PR #160, opened 2024-09-06
DM-46167: added optical configuration to enums

parfa30:
While we wait for the XML changes to be made, we will include the Tunable Laser Optical Configuration here. It will be removed when the xml has been deployed.


tribeiro (approved):
Looks good.. please, rebase to develop before merging.

PR #161, opened 2024-09-06
Add unpark MTDome logic for MTCS

iglesu:
Add unpark MTDome logic for MTCS and related tests
    
    - Implemented unpark logic in MTCS to move the dome from PARKED state.
    - Added mock support for dome events in MTCSAsyncMock.
    - Included unit tests for unpark_dome


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -668,6 +715,58 @@ async def park_dome(self) -> None:
...
+        az_motion = await self.rem.mtdome.evt_azMotion.aget(timeout=self.fast_timeout)
+
+        if az_motion.state == MTDome.MotionState.PARKED:
+            self.log.info("Dome is currently PARKED. Proceeding to un-park.")
+
+            current_position = await self.rem.mtdome.tel_azimuth.aget(
+                timeout=self.fast_timeout
+            )
+
+            az_delta = 0.1  # A small move to un-park the Dome
```
tribeiro, line 735: Please make this a class attribute instead.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -25,6 +25,7 @@
 import copy
 import enum
 import logging
+import time
```
tribeiro, line 28: Please, don't use the `time` module, you already have timing functionality with the `lsst.ts.utils` module. You can user `utils.current_tai()`.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -632,6 +633,52 @@ def dome_ready(az_motion: salobj.type_hints.BaseMsgType) -> bool:
                 f"Dome state: {MTDome.MotionState(az_motion.state).name}, inPosition: {az_motion.inPosition}"
             )
 
+    async def wait_for_dome_azEnabled(self, timeout: float) -> None:
```
tribeiro, line 636: This method name should be `snake_case`, please rename it to `wait_for_dome_az_enabled`.

However, do you really need this method? won't the one you included for parking (`wait_for_dome_state`) do the work?

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -668,6 +715,58 @@ async def park_dome(self) -> None:
...
+            self.rem.mtdome.evt_azMotion.flush()
+
+            await self.rem.mtdome.cmd_moveAz.set_start(
+                position=unparked_position,
+                velocity=unparked_velocity,
+                timeout=self.park_dome_timeout,
+            )
+
+            # The cmd_MoveAz should issue a azEnabled event set to ENABLED
+            await self.wait_for_dome_azEnabled(timeout=self.park_dome_timeout)
```
tribeiro, line 749: So you really need to look for this event? Why  not rely on the `wait_for_dome_state`?

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -121,10 +121,14 @@ async def setup_types(self) -> None:
         )
 
         # MTDome Motion PARKED state. State is set by the mocked park cmd.
-        self._mtdome_evt_azMotion_state = types.SimpleNamespace(
+        self._mtdome_evt_azMotion = types.SimpleNamespace(
```
tribeiro, line 124: Please, use `snake_case` for attributes.

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
```diff
@@ -121,10 +121,14 @@ async def setup_types(self) -> None:
...
             state=MTDome.MotionState.UNDETERMINED, inPosition=False
         )
 
+        self._mtdome_evt_azEnabled = types.SimpleNamespace(
```
tribeiro, line 128: `snake_case`

tribeiro (changes requested):
I some some questions regarding the use of `azEnabled` event, why not rely on the `azMotion` like we did for parking?

Also, please, make sure all methods and attributes are `snake_case`.

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: on the azEnabled event, I'm following the advice given by you and Wouter on the issue: https://rubinobs.atlassian.net/browse/DM-45608?focusedCommentId=523258 specifically:

`Also check the azMotion, and azEnabled events to verify UNPARKED status.`


but happy to do otherwise.

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: shall we also add the
unpark_velocity = 0.0   # The Dome defaults to a crawl velocity

as an attribute?

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: as per @tribeiro comment: 
`velocity will always be zero, no matter what.. you can even ignore that parameter and don't set it as the default is zero.`

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: I removed the azEnabled check

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: azEnabled check removed

python/lsst/ts/observatory/control/maintel/mtcs.py
iglesu: as per @tribeiro comment: 
`velocity will always be zero, no matter what.. you can even ignore that parameter and don't set it as the default is zero.`

python/lsst/ts/observatory/control/mock/mtcs_async_mock.py
iglesu: I'm removing the azEnabled check

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -148,6 +148,9 @@ def __init__(
         self._dome_az_in_position: typing.Union[None, asyncio.Event] = None
         self._dome_el_in_positio: typing.Union[None, asyncio.Event] = None
 
+        self.az_delta = 0.1  # A small move to un-park the Dome
```
tribeiro, line 151: so, instead of relying on the comment (that is only really here) to specify what the attribute is, use a more expressive name. Something like;

```
# Small offset (in degrees) applied to the dome to unpark it.
self.dome_az_unpark_offset = 0.1
```

Would be better.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -148,6 +148,9 @@ def __init__(
...
+        self.dome_unpark_velocity = 0.0  # A small move to un-park the Dome
```
tribeiro, line 152: This is not necessary so please, consider removing it.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -668,6 +671,53 @@ async def park_dome(self) -> None:
...
             )
 
+    async def unpark_dome(self) -> None:
+        """Un-Park the dome by moving it a small delta amount."""
+
+        await self.assert_all_enabled(
+            message="All components need to be enabled for un-parking the Dome."
+        )
+
+        self.log.info("Checking if the dome is currently PARKED.")
```
tribeiro, line 681: Maybe debug level instead?

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -668,6 +671,53 @@ async def park_dome(self) -> None:
...
+            current_position = await self.rem.mtdome.tel_azimuth.aget(
+                timeout=self.fast_timeout
+            )
+
+            unparked_position = current_position.positionActual + self.az_delta
+
+            self.rem.mtdome.evt_azMotion.flush()
+
+            # We don't specify the dome velocity as it defaults to a
+            # `crawl` which is what we want.
```
tribeiro, line 696: This is not correctly stated. This should read:

```
# We don't specify the dome velocity as it defaults to
# zero, which is what we want.
```

tribeiro (approved):
looks good!

PR #164, opened 2024-09-12
DM-46276: modified wavelength change and setup laser

parfa30:
Changing the laser wavelength and setup laser functions in mtcalsys so are more usable in a variety of sal scripts.

### DIFF
--- a/doc/news/DM-46276.feature.rst
+++ b/doc/news/DM-46276.feature.rst
@@ -1,0 +1,8 @@
+Add features to allow ``MTCalSys`` to better handle the laser
+- In ``mtcalsys.py`` made the following changes: 
+ - Added ``laser_start_propagate`` and ``laser_stop_propagate()``
+ - Added ``get_laser_parameters()``
+ - Improved ``setup_laser()`` to change the wavelength and the optical configuration
+ - Changed ``change_laser_wavelength()`` so it can be used for the laser or whitelight system 
+- In ``mtcalsys.yaml`` added a laser functional setup
+- In ``mtcalsys_schema.yaml`` added laser mode and optical configuration
--- a/python/lsst/ts/observatory/control/maintel/mtcalsys.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcalsys.py
@@ -27,6 +27,9 @@ from dataclasses import dataclass
 
 import numpy as np
 from lsst.ts import salobj, utils
+
+# TODO: (DM-46168) Revert workaround for TunableLaser XML changes
+from lsst.ts.observatory.control.utils.enums import LaserOpticalConfiguration
 from lsst.ts.xml.enums.TunableLaser import LaserDetailedState
 
 from ..base_calsys import BaseCalsys
@@ -157,6 +160,7 @@ class MTCalsys(BaseCalsys):
         self.led_rest_position = 120.0  # mm
 
         self.laser_enclosure_temp = 20.0  # C
+        self.laser_warmup = 20.0  # sec
 
         self.exptime_dict: dict[str, float] = dict(
             camera=0.0,
@@ -164,45 +168,8 @@ class MTCalsys(BaseCalsys):
             fiberspectrograph=0.0,
         )
 
-    def calculate_laser_focus_location(self, wavelength: float) -> float:
-        """Calculates the location of the linear stage that provides the focus
-        for the laser projector. This location is dependent on the
-        wavelength of the laser.
-
-        Parameters
-        ----------
-        wavelength : `float`
-            wavelength of the laser projector in nm
-
-        Returns
-        ------
-        location of the linear stage for the laser projector focus in mm
-
-        """
-        # TODO (DM-44772): implement the actual function
-        return 10.0
-
-    async def change_laser_wavelength(self, wavelength: float) -> None:
-        """Change the TunableLaser wavelength setting
-
-        Parameters
-        ----------
-        wavelength : `float`
-            wavelength of the laser in nm
-        """
-
-        task_wavelength = self.rem.tunablelaser.cmd_changeWavelength.set_start(
-            wavelength=wavelength, timeout=self.long_long_timeout
-        )
-        task_focus = self.linearstage_laser_focus.cmd_moveAbsolute.set_start(
-            distance=self.calculate_laser_focus_location(wavelength),
-            timeout=self.long_long_timeout,
-        )
-
-        await asyncio.gather(task_wavelength, task_focus)
-
     async def is_ready_for_flats(self) -> bool:
-        """Designates if the calibraiton hardware is in a state
+        """Designates if the calibration hardware is in a state
         to take flats.
         """
         # TODO (DM-44310): Implement method to check that the
@@ -240,12 +207,94 @@ class MTCalsys(BaseCalsys):
             await self.linearstage_projector_select.cmd_moveAbsolute.set_start(
                 distance=self.ls_select_laser_location, timeout=self.long_timeout
             )
-            await self.setup_laser(config_data["laser_mode"])
-            await self.rem.tunablelaser.cmd_startPropagating.start(
-                timeout=self.long_long_timeout
+            await self.setup_laser(
+                config_data["laser_mode"],
+                config_data["wavelength"],
+                config_data["optical_configuration"],
+            )
+            await self.laser_start_propagate()
+
+    def calculate_laser_focus_location(self, wavelength: float = 500.0) -> float:
+        """Calculates the location of the linear stage that provides the focus
+        for the laser projector. This location is dependent on the
+        wavelength of the laser.
+
+        Parameters
+        ----------
+        wavelength : `float`
+            wavelength of the laser projector in nm
+            Default 500.0
+
+        Returns
+        ------
+        location of the linear stage for the laser projector focus in mm
+
+        """
+        # TODO (DM-44772): implement the actual function
+        return 10.0
+
+    async def change_laser_wavelength(
+        self,
+        wavelength: float,
+        use_projector: bool = True,
+    ) -> None:
+        """Change the TunableLaser wavelength setting
+
+        Parameters
+        ----------
+        wavelength : `float`
+            wavelength of the laser in nm
+        use_projector : `bool`
+            identifies if you are using the projector while
+            changing the wavelength.
+            Default True
+        """
+        task_wavelength = self.rem.tunablelaser.cmd_changeWavelength.set_start(
+            wavelength=wavelength, timeout=self.long_long_timeout
+        )
+
+        if use_projector:
+            task_focus = self.linearstage_laser_focus.cmd_moveAbsolute.set_start(
+                distance=self.calculate_laser_focus_location(wavelength),
+                timeout=self.long_long_timeout,
+            )
+            await asyncio.gather(task_wavelength, task_focus)
+
+        else:
+            await task_wavelength
+
+    async def change_laser_optical_configuration(
+        self, optical_configuration: LaserOpticalConfiguration
+    ) -> None:
+        """Change the output of the laser.
+
+        Parameters
+        ----------
+        optical_configuration : LaserOpticalConfiguration
+        """
+        assert optical_configuration in list(LaserOpticalConfiguration)
+
+        current_configuration = (
+            await self.rem.tunablelaser.evt_opticalConfiguration.aget()
+        )
+        if current_configuration.configuration != optical_configuration:
+            self.log.debug(
+                f"Changing optical configuration from {current_configuration} to {optical_configuration}"
+            )
+            await self.rem.tunablelaser.cmd_setOpticalConfiguration.set_start(
+                configuration=optical_configuration, timeout=self.long_timeout
             )
 
-    async def setup_laser(self, mode: LaserDetailedState) -> None:
+        else:
+            self.log.debug("Laser Optical Configuration already in place.")
+
+    async def setup_laser(
+        self,
+        mode: LaserDetailedState,
+        wavelength: float,
+        optical_configuration: LaserOpticalConfiguration = LaserOpticalConfiguration.SCU,
+        use_projector: bool = True,
+    ) -> None:
         """Perform all steps for preparing the laser for monochromatic flats.
         This includes confirming that the thermal system is
         turned on and set at the right temperature. It also checks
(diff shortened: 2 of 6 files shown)


python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
-            await self.rem.tunablelaser.cmd_startPropagating.start(
-                timeout=self.long_long_timeout
+            await self.setup_laser(
+                config_data["laser_mode"],
+                config_data["wavelength"],
+                config_data["optical_configuration"],
+            )
+            await self.laser_start_propagate()
+
+    def calculate_laser_focus_location(self, wavelength: float | None) -> float:
```
edennihy, line 217: Do really you want to have a`None` option here? This method hasn't been implemented yet, but relates to the next few comments.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+        location of the linear stage for the laser projector focus in mm
+
+        """
+        # TODO (DM-44772): implement the actual function
+        return 10.0
+
+    async def change_laser_wavelength(
+        self,
+        wavelength: float | None,
+        use_projector: typing.Optional[bool] = True,
```
edennihy, line 238: Since this is a bool value with a default, you don't actually need to make it Optional. You can just use `use_projector: bool = True`. The typing.Option[bool] is actually equivalent to the way you used e.g. `bool | None` above.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+        use_projector : `bool`
+            identifies if you are using the projector while
+            changing the wavelength
+        """
+        if wavelength is not None:
+            task_wavelength = self.rem.tunablelaser.cmd_changeWavelength.set_start(
+                wavelength=wavelength, timeout=self.long_long_timeout
+            )
+
+        if use_projector:
```
edennihy, line 255: If you want to keep the option to update the focus location without updating the laser wavelength by passing `wavelength=None`, we should consider splitting this into a separate method.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+        optical_configuration : LaserOpticalConfiguration
+        """
+        current_configuration = (
+            await self.rem.tunablelaser.evt_opticalConfiguration.aget().configuration
+        )
+        if current_configuration != optical_configuration:
+            self.log.debug(
+                f"Changing optical configuration from {current_configuration} to {optical_configuration}"
+            )
+            if optical_configuration in list(LaserOpticalConfiguration):
```
edennihy, line 281: Since we already enforced that the `optical_configuration` is a LaserOpticalConfiguration in parameter definition, you don't actually need to perform this check here. If `optical_configuration` is not in the list when the user calls this method, it will fail.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+                await self.rem.tunablelaser.cmd_setOpticalConfiguration.set_start(
```
edennihy, line 282: If the `optical_configuration` being passed to the method is a proper enumeration like you enforced, this is the only command you need to pull out of this loop.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+                    configuration=optical_configuration, timeout=self.long_timeout
+                )
+            elif LaserOpticalConfiguration[optical_configuration] in list(
```
edennihy, line 285: This whole elif should be removed since you are already enforcing that `optical_configuration` is an enumeration and not a string.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+                LaserOpticalConfiguration
+            ):
+                await self.rem.tunablelaser.cmd_setOpticalConfiguration.set_start(
+                    configuration=LaserOpticalConfiguration[optical_configuration],
+                    timeout=self.long_timeout,
+                )
+            else:
```
edennihy, line 292: Since this already suggested by the parameter definition, you can also drop this.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+                raise RuntimeError(
```
edennihy, line 293: If you really want to ensure that the input is a valid enumeration, you could, at the top of the method add a line:
`assert optica_configuration in LaserOpticalConfiguration`, and then avoid the if/else loops all together. That would also allow users to pass a string as long as it's a valid enumeration key.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+                    f"{optical_configuration} not an acceptable LaserOpticalConfiguration"
+                )
+
+        else:
+            self.log.debug("Laser Optical Configuration already in place.")
+
+    async def setup_laser(
+        self,
+        mode: LaserDetailedState,
+        wavelength: typing.Optional[float] = 500.0,
```
edennihy, line 303: This does not need to be optional since the type should always be float. Can instead use:
`float = 500.0` to set it as a float with a default 500.0

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+        optical_configuration: typing.Optional[
```
edennihy, line 304: Same thing here, no need for this to optional since you are only allowing one type.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -240,12 +207,105 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+            LaserOpticalConfiguration
+        ] = LaserOpticalConfiguration.SCU,
+        use_projector: typing.Optional[bool] = True,
```
edennihy, line 307: Finally, same thing here. Can just be `bool = True'

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -255,8 +315,14 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
         ----------
         mode : LaserDetailedState
             Mode of the TunableLaser
-            Options: CONTINUOUS, BURST, TRIGGER
-
+            Options: CONTINUOUS, BURST
+        wavelength : `float`
+            Wavelength fo the laser in nm
```
edennihy, line 320: Please add the default value to the docstring here.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -255,8 +315,14 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
...
+        optical_configuration : LaserOpticalConfiguration
+            Output of laser
```
edennihy, line 322: Please add the default value to the docstring here.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -255,8 +315,14 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
...
+        use_projector : `bool`
+            identifies if you are using the projector while
+            changing the wavelength
```
edennihy, line 325: Please add the default value to the docstring here.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -279,6 +345,24 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
...
             )
 
+        await self.change_laser_optical_configuration(optical_configuration)
+        await self.change_laser_wavelength(wavelength, use_projector)
+
+    async def laser_start_propagate(self) -> None:
+        """Start the propagation of the Tunable Laser"""
+        # get laser detailed state
+
+        await self.rem.tunablelaser.cmd_startPropagateLaser.start(
```
edennihy, line 355: How does the propagation work? This might be a good situation to add a check on a evt to check that the propagation has actually started. You can check against the detailedstate evt that the laser is propagating with the right mode before this method completes. 

Maybe using something like this https://github.com/lsst-ts/ts_observatory_control/blob/cab4bdbae2dbf41863a87eadc272396a523dc9e5/python/lsst/ts/observatory/control/auxtel/atcalsys.py#L675

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -279,6 +345,24 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
...
+            timeout=self.laser_warmup
+        )
+
+    async def laser_stop_propagate(self) -> None:
+        """Stop the propagation of the Tunable Laser"""
+
+        await self.rem.tunablelaser.cmd_stopPropagateLaser.start(
+            timeout=self.long_timeout
+        )
+
```
edennihy, line 365: Same thing here, you should add a check that the detailed state is nonpropagating before this method completes.

edennihy (changes requested):
Hi Parker, I added a handful of recommended changes and suggestions to start here. @tribeiro actually looked at most of them together to make sure we were in agreement, so we should be pretty close to finished once these changes are in. 

One other missing piece is the unit tests for these methods. The nice part about that though is since they are being unit tested here, you will not need to test them again in the scripts when they are used.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -279,6 +345,24 @@ async def setup_laser(self, mode: LaserDetailedState) -> None:
                 f"{mode} not an acceptable LaserDetailedState [CONTINOUS, BURST, TRIGGER]"
             )
 
+        await self.change_laser_optical_configuration(optical_configuration)
+        await self.change_laser_wavelength(wavelength, use_projector)
+
+    async def laser_start_propagate(self) -> None:
+        """Start the propagation of the Tunable Laser"""
+        # get laser detailed state
```
edennihy, line 353: I think you can remove this comment

doc/news/DM-46276.feature.2.rst
edennihy, line 1: The features index from zero, so this file should actually be named feature.1.rst

doc/news/DM-46276.feature.rst
edennihy, line 1: Could you be a bit more specific in a few bullet points about the changes here? We could for example list the new methods and a quick description of they did. Here is an example of how to format this in the rst file https://github.com/lsst-ts/ts_observatory_control/blob/develop/doc/news/DM-46458.feature.rst


Add feature to allow \``ATCalSys\`` to skip monochromator configuration. 

\- In \``atcalsys_schema.yaml\``, add default values for wavelength, entrace_slit and exit_slit. 
  Add option to set monochromator_grating to None to skip monchromator configuration and set None as default value.
\- In \``atcalsys.py\``, add feature to skip configuring monochromator if monchromator_grating is None.
\- In \``atcalsys.yaml\``, update monochromator configuration values for ptc curves to skip monchromator configuration.

Which then displays like this:
Add feature to allow ``ATCalSys`` to skip monochromator configuration. 

- In ``atcalsys_schema.yaml``, add default values for wavelength, entrace_slit and exit_slit. 
  Add option to set monochromator_grating to None to skip monchromator configuration and set None as default value.
- In ``atcalsys.py``, add feature to skip configuring monochromator if monchromator_grating is None.
- In ``atcalsys.yaml``, update monochromator configuration values for ptc curves to skip monchromator configuration.

edennihy (commented):
Thanks for the changes, please see my final comments to help improve the docnotes. Otherwise, you can feel free to squash the commits and prepare to merge. 

The last thing we need are a few unit tests for these methods. You could follow the examples here for how to setup a set of basic tests for these methods:

Example method -> https://github.com/lsst-ts/ts_observatory_control/blob/30f7d56a0b6eeabf7c60ffa2e6c247cbf90b0820/python/lsst/ts/observatory/control/maintel/mtcs.py#L199

Example Unit Test -> https://github.com/lsst-ts/ts_observatory_control/blob/30f7d56a0b6eeabf7c60ffa2e6c247cbf90b0820/tests/maintel/test_mtcs.py#L544

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -342,6 +340,25 @@ async def setup_laser(
...
+        -------
+            list : configuration details
+
+        """
+
+        return await asyncio.gather(
+            self.rem.tunablelaser.evt_opticalConfiguration.aget(
+                timeout=self.long_timeout
+            ),
+            self.rem.tunableevt_wavelengthChanged.aget(timeout=self.long_timeout),
```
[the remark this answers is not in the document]
edennihy: This comment still needs to be resolved.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -342,6 +340,25 @@ async def setup_laser(
...
+            self.rem.tunableevt_interlockState.aget(timeout=self.long_timeout),
```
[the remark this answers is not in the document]
edennihy: Still needs to be resolved.

edennihy (approved):
Looks good, thanks for all the changes!

PR #165, opened 2024-09-18
tickets/DM-45821: Implement MT Dome slewing in MTCS

cvillalon:


tribeiro (approved):
Looks good!

cvillalon:
I've pushed some mock definitions that were missing and caused the hanging in `test_slew_dome_to`. I think it is solved now. You may want to take another look @tribeiro so I'll wait for your confirmation before merging.

PR #170, opened 2024-10-11
DM-46179 Extend TCS readiness check to other image types beyond OBJECT

iglesu:


tests/test_generic_camera.py
```diff
@@ -316,3 +317,125 @@ async def test_take_cwfs(self) -> None:
...
+    async def test_take_image_no_tcs_sync(self) -> None:
+        image_types_no_tcs = ["BIAS", "DARK", "FLAT", "FOCUS"]
+
+        for img_type in image_types_no_tcs:
+            with self.subTest(image_type=img_type):
+                with self.get_fake_tcs() as fake_tcs:
+                    fake_tcs.fail = False
+                    methods = self.image_types.get(img_type)
+                    if not methods:
+                        self.fail(
```
tribeiro, line 370: Why not just raise an exception? You could do something like

```
assert not methods, f"Image type '{img_type}' not found in image_types mapping."
```

tests/test_generic_camera.py
```diff
@@ -316,3 +317,125 @@ async def test_take_cwfs(self) -> None:
...
+                        )
+
+                    assert_method_name = methods["assert_method"]
+                    params = methods["params"]
+
+                    self.reset_mocks()
+
+                    assert_take_method = getattr(self, assert_method_name, None)
+                    if assert_take_method is None:
+                        self.fail(
```
tribeiro, line 381: See comment above.

tests/test_generic_camera.py
```diff
@@ -316,3 +317,125 @@ async def test_take_cwfs(self) -> None:
...
+                    assert fake_tcs.called == 0
+
+    async def test_take_image_tcs_sync(self) -> None:
+        image_types_tcs = ["OBJECT", "ENGTEST", "ACQ", "CWFS"]
+
+        for img_type in image_types_tcs:
+            with self.subTest(image_type=img_type):
+                methods = self.image_types.get(img_type)
+                if not methods:
+                    self.fail(
```
tribeiro, line 397: See comment above.

tests/test_generic_camera.py
```diff
@@ -316,3 +317,125 @@ async def test_take_cwfs(self) -> None:
...
+                )
+
+    async def test_take_image_tcs_sync_fail(self) -> None:
+        image_types_tcs = ["OBJECT", "ENGTEST", "ACQ", "CWFS"]
+
+        for img_type in image_types_tcs:
+            with self.subTest(image_type=img_type):
+                methods = self.image_types.get(img_type)
+                if not methods:
+                    self.fail(
```
tribeiro, line 418: See comment above

tribeiro (commented):
It mostly look ok but I do have some comments about how you did the unit tests.

tests/test_generic_camera.py
iglesu: ok, let me try that, I was trying to get around a mypy issue, that was thrown on the method signature type.

python/lsst/ts/observatory/control/base_camera.py
```diff
@@ -965,7 +965,10 @@ async def take_imgtype(
         if imgtype not in ["BIAS", "DARK"]:
             await self.setup_instrument(**kwargs)
 
-        if imgtype == "OBJECT" and self.ready_to_take_data is not None:
+        if (
+            imgtype in ["OBJECT", "ENGTEST", "ACQ", "CWFS"]
+            and self.ready_to_take_data is not None
+        ):
```
tribeiro, line 971: A couple line below (line 983) there is another `elif` that needs updating. It warns users that the TCS synchronization feature is not configured. You might want to extract the list into a variable to use in both places, e.g.:

```
sync_imgtypes = ["OBJECT", "ENGTEST", "ACQ", "CWFS"]

if imgtype in sync_imgtypes:
    ...
```

tribeiro (commented):
I only have a few inline comments. Also, please add a second news fragment for the idl -> xml updates you did and for the fix to the MTRotator enumeration. I think the first one can be added in a file called `DM-46179.removal.rst` and the second one to `DM-46179.bugfix.rst`.

Also, I am ready to approve the PR after you make these updates so feel free to re-organize the git history.

tribeiro (approved):
Looks good!

PR #174, opened 2024-10-28
tickets/DM-46980: Implement dome homing in MTCS

cvillalon:
This implementation encodes the manual procedure to home the MTDome azimuth including the case where the reference azimuth position is lost, and the encoders are not reading the real position of the dome azimuth.

### DIFF
--- a/python/lsst/ts/observatory/control/maintel/mtcs.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcs.py
@@ -141,6 +141,7 @@ class MTCS(BaseTCS):
         self.dome_flat_az = 20.0
         self.dome_flat_el = self.dome_park_el
         self.dome_slew_tolerance = Angle(1.5 * u.deg)
+        self.home_dome_az = 328.0
 
         # TODO (DM-45609): This is an initial guess for the time it takes the
         #  dome to park. It might need updating.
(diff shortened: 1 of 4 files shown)


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -841,9 +842,40 @@ async def close_m1_cover(self) -> None:
...
-        raise NotImplementedError("# TODO: Implement (DM-21336).")
+    async def home_dome(self, physical_az: float = 0.0) -> None:
+        self.log.info("Homing dome")
+        reported_az = await self.rem.mtdome.tel_azimuth.aget(timeout=self.fast_timeout)
+
+        offset = physical_az - reported_az.positionActual
+        self.log.debug(f"Dome azimuth offset: {offset} degrees")
+        target_az = Angle(self.home_dome_az - offset, unit=u.deg).deg
+        self.rem.mtdome.evt_azMotion.flush()
+        await self.rem.mtdome.cmd_moveAz.set_start(
```
tribeiro, line 853: The MTCS class has a `slew_dome_to` method that will handle all this for you.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -841,9 +842,40 @@ async def close_m1_cover(self) -> None:
...
+
+        await self._handle_in_position(
+            self.rem.mtdome.evt_azMotion,
+            timeout=self.fast_timeout,
+            settle_time=self.tel_settle_time,
+            component_name="MTDome",
+        )
+
+        await self.rem.mtdome.cmd_stop.set_start(
+            engageBrakes=True, timeout=self.long_long_timeout
```
tribeiro, line 865: you need to pass a `subSystemIds`. There's an enumeration for this you can use;

```
from lsst.ts.xml.enums.MTDome import SubSystemId
```

then;

```
      await self.rem.mtdome.cmd_stop.set_start(
            engageBrakes=True, subSystemIds =SubSystemId.AMCS, timeout=self.long_long_timeout
       )
```

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -841,9 +842,40 @@ async def close_m1_cover(self) -> None:
                 f"{MTMount.DeployableMotionState.DEPLOYED!r}"
             )
 
-    async def home_dome(self) -> None:
-        # TODO: Implement (DM-21336).
-        raise NotImplementedError("# TODO: Implement (DM-21336).")
+    async def home_dome(self, physical_az: float = 0.0) -> None:
```
tribeiro, line 845: Please add docstring and document the input parameter.

tribeiro (commented):
Here some initial comments.. look pretty good, just some minor comments...

python/lsst/ts/observatory/control/maintel/mtcs.py
cvillalon: You're absolutely right! It slipped through my mind even though I think I wrote that one.

tribeiro: Let's use the default 120s for now.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -841,9 +842,40 @@ async def close_m1_cover(self) -> None:
...
+            engageBrakes=True,
+            subSystemIds=MTDome.SubSystemId.AMCS,
+            timeout=self.long_long_timeout,
+        )
+        motion_state = await self.rem.mtdome.evt_azMotion.aget(
+            timeout=self.fast_timeout
+        )
+        while motion_state.state != MTDome.MotionState.STOPPED_BRAKED:
+            motion_state = await self.rem.mtdome.evt_azMotion.next(
+                flush=False, timeout=self.fast_timeout
```
tribeiro, line 872: I think `fast_timeout` is too short here.. you probably want something like `long_long_timeout`

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -841,9 +842,40 @@ async def close_m1_cover(self) -> None:
...
+        self.log.debug(f"Dome azimuth offset: {offset} degrees")
+        target_az = self.home_dome_az - offset
+        await self.slew_dome_to(target_az)
+
+        await self.rem.mtdome.cmd_stop.set_start(
+            engageBrakes=True,
+            subSystemIds=MTDome.SubSystemId.AMCS,
+            timeout=self.long_long_timeout,
+        )
+        motion_state = await self.rem.mtdome.evt_azMotion.aget(
```
tribeiro, line 867: Make sure you flush this event before you call `cmd_stop`

tribeiro (commented):
A couple more comments.. also I think you need to rebase this.

PR #177, opened 2024-11-11
Tickets/DM-47497 Adding CBP to MTCalsys

elanaku:


python/lsst/ts/observatory/control/data/mtcalsys.yaml
```diff
@@ -87,3 +87,20 @@ laser_functional:
   use_fiberspectrograph_blue: false
   exposure_times:
     - 15.0
+
+cbp_calibration:
+  calib_type: CBPCalibration
```
parfa30, line 92: this is defined as an enum in `ts_observatorycontrol/utils/enums.py`. `calib_type` identifies the light source. So let's keep this as `Mono` and introduce a new variable called `use_cbp` or something like that.

python/lsst/ts/observatory/control/data/mtcalsys.yaml
```diff
@@ -87,3 +87,20 @@ laser_functional:
...
+  use_camera: false
+  wavelength: 700
+  wavelength_width: 800
+  wavelength_resolution: 50
+  use_cbpelectrometer: true
```
parfa30, line 97: I suppose we should rename the other electrometer to be flatfieldelectrometer or something like that.

parfa30: Also, `use_electrometer` is required currently. You'll have to add these two new ones to mtcalsys_schema.yaml

python/lsst/ts/observatory/control/data/mtcalsys.yaml
```diff
@@ -87,3 +87,20 @@ laser_functional:
...
+  use_cbpelectrometer: true 
+  use_cbpcalelectrometer: true
+  use_fiberspectrograph_red: false
+  use_fiberspectrograph_blue: false
+  electrometer_integration_time: 0.001
+  electrometer_mode: CHARGE
+  electrometer_range: 2e-6
+  laser_mode: BURST
+  optical_configuration: F1_SCU
+  exposure_times: 30
```
parfa30, line 106: I don't know it it will be unhappy unless you put it in this format. It might be fine.
exposure_times: 
 - 30.0

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -200,6 +206,27 @@ async def setup_calsys(self, sequence_name: str) -> None:
             await self.linearstage_projector_select.cmd_moveAbsolute.set_start(
                 distance=self.ls_select_led_location, timeout=self.long_timeout
             )
+        elif (calibration_type == CalibrationType.CBPCalibration) or (
+            calibration_type == CalibrationType.CBP
```
parfa30, line 210: As I mentioned above, I thinks CalibrationType should remain the "light input" and we should create a new variable for the CBP vs. FlatField

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -200,6 +206,27 @@ async def setup_calsys(self, sequence_name: str) -> None:
...
+        ):
+            await self.setup_cbp(
+                config_data["cbp_azimuth"],
+                config_data["cbp_elevation"],
+                config_data["cbp_mask"],
+                config_data["cbp_focus"],
+                config_data["cbp_rotation"],
```
parfa30, line 217: Do you want to add this to the mtcalsys.yaml file?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +366,47 @@ async def setup_laser(
...
+        el : `float`
+            Elevation of CBP in degrees
+        mask : `int`
+            Mask number to use
+        focus: `float`
+            Focus position in um
+        rot: `int`
+            Rotator position of mask in degrees
+            Default 0
+        use_projector : `bool`
```
parfa30, line 392: You can remove this.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +366,47 @@ async def setup_laser(
         await self.change_laser_optical_configuration(optical_configuration)
         await self.change_laser_wavelength(wavelength, use_projector)
 
+    async def setup_cbp(
+        self,
+        azimuth: float,
+        elevation: float,
+        mask: typing.Optional[int] = None,
```
parfa30, line 373: Do we want this to be optional or have a default here? I think every time we move the CBP it would be good to confirm the mask, focus and rotation.

parfa30 (commented):
After reading through all this code and thinking about things a bit more, I think we should add CBP to `calib_type` but not include CBPCalibration. That can be its own test in mtcalsys.yaml, but I don't think is a whole separate calib_type.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +361,43 @@ async def setup_laser(
...
+            Elevation of CBP in degrees
+        mask : `int`
+            Mask number to use
+        focus: `float`
+            Focus position in um
+        rot: `int`
+            Rotator position of mask in degrees
+            Default 0
+        """
+        timeout = 60
```
tribeiro, line 388: Please, don't use local timeout variable, we define some [default timeouts](https://github.com/lsst-ts/ts_observatory_control/blob/09ee0a6aac516108684cc77ba0d8630c6c0a3d56/python/lsst/ts/observatory/control/remote_group.py#L260-L262) in the base class. Chose one of those or make a custom one at the class level.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -751,6 +860,7 @@ async def take_electrometer_scan(
         self,
         exposure_time: float | None,
         exposures_done: asyncio.Future,
+        electrometer_to_scan: salobj,
```
tribeiro, line 863: salobj is a module, not a type. What is the type of this parameter?

PR #179, opened 2024-11-14
tickets/DM-47223:  Prepare _wait_hard_point_test_ok method in MTCS for parallelization

cvillalon:
Update the wait method for hardpoint tests to be ok in order to make it compatible with concurrent executions.
    
Using `next` will pop the data from the queue so it might happen that one execution will “steal” the event for the other and vice versa. This PR changes `next` to `aget` and adds a `sleep` to avoid that.


doc/news/DM-47223.feature.rst
```diff
@@ -0,0 +1,2 @@
+Update ``_wait_hard_point_test_ok`` method in ``MTCS`` to be compatible with
+concurrent executions.
```
tribeiro, line 2: For documentation in general we adopted something called "semantic line break". Basically it means that each phrase should be in one line. Please, make sure this is all in one line.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1285,10 +1285,13 @@ async def _wait_hard_point_test_ok(self, hp: int) -> None:
...
-        while True:
+        timer_task = asyncio.create_task(
+            asyncio.sleep(self.timeout_hardpoint_test_status)
+        )
+        while not timer_task.done():
             hp_test_state = MTM1M3.HardpointTest(
                 (
-                    await self.rem.mtm1m3.evt_hardpointTestStatus.next(
+                    await self.rem.mtm1m3.evt_hardpointTestStatus.aget(
                         flush=False, timeout=self.timeout_hardpoint_test_status
```
tribeiro, line 1295: `aget` does not have a `flush` parameter.

python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1302,6 +1305,10 @@ async def _wait_hard_point_test_ok(self, hp: int) -> None:
             else:
                 self.log.info(f"Hard point {hp} test state: {hp_test_state!r}.")
 
+            await self.rem.mtm1m3.evt_heartbeat.next(
```
tribeiro, line 1308: you might want to wrap this await here with a `try/except` and catch and treat timeout errors. It might be hard for users to understand why there is a heartbeat timeout coming from this loop. Consider something like:

```
try:
    await self.rem.mtm1m3.evt_heartbeat.next(...)
expect asyncio.TimeoutError:
    raise RuntimeError(f"No heartbeat received from M1M3 in the last {self.timeout_hardpoint_test_status}s while waiting for hard point data information. Check CSC liveliness.")
```

tribeiro (commented):
I have some small comments I hope you can address before I can approve it.

tribeiro (approved):

PR #181, opened 2024-11-25
Tickets/dm-47795

kfanning:
Fixing OBS-708 - `m1m3/check_actuators.py` exits M1M3 engineering mode before the secondary test of the last actuator if the primary fails. This change causes `mtcs._wait_bump_test_ok` to wait for both tests to finish before raising a RuntimeError in the case one fails.


tribeiro (changes requested):
Not sure I understand what is going on here. Can you please clarify what it is you are trying to fix?

There's no description in the ticket of what you are trying to fix and the context here does not allow me to understand the problem.

edennihy:
This feature was added as part of https://rubinobs.atlassian.net/browse/OSW-494. Will close PR.

PR #185, opened 2024-12-09
Tickets/DM-47619: Add support to `RemoteGroup` for disabling checks for a list of components

MarcoRocchietti:


python/lsst/ts/observatory/control/remote_group.py
```diff
@@ -317,6 +317,26 @@ def __init__(
             asyncio.gather(*start_task_list) if len(start_task_list) > 0 else None
         )
 
+    @staticmethod
+    def _remote_name_to_attr_format(salobj_remote_name: str) -> str:
```
tribeiro, line 321: consider renaming `salobj_remote_name` -> `component_name`.

tribeiro (approved):
looks good, I only have one comment about the argument name on that `_remote_name_to_attr_format`.

PR #186, opened 2024-12-09
DM-48023: Offset script shall enable and then disable M1M3 slew flag.

tribeiro:


dsanmartim (approved):
Sorry, I took some time to make sure I could understand the change. It looks good.

PR #191, opened 2025-02-25
DM-49065: get_projector_status function

parfa30:
This function reads out the state of the linear stages and LED controller in the calibration projector

### DIFF
--- a/python/lsst/ts/observatory/control/maintel/mtcalsys.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcalsys.py
@@ -395,8 +395,20 @@ class MTCalsys(BaseCalsys):
 
     async def get_projector_setup(
         self,
-    ) -> None:
-        """Get configuration of flatfield projector"""
+    ) -> tuple[str, float, float, float, str]:
+        """Get configuration of flatfield projector largely for
+        debugging purposes
+
+        Return
+        ------
+            tuple:
+                projector_location: whether or not the vertical stage is
+                aligned with led or laser output or neither
+                led_location: which LED system is aligned with teh optical path
+                led_focus: value of the led focus stage
+                laser_focus: value of the laser focus stage
+                led_state: shows whether LEDs are ON/OFF
+        """
 
         select_location = await self.linearstage_projector_select.tel_position.next(
             flush=True
@@ -424,6 +436,14 @@ class MTCalsys(BaseCalsys):
             f"LED Focus stage pos @: {led_focus.position}, \n"
             f"Laser Focus stage pos @: {laser_focus.position}, \n"
             f"LED State stage pos @: {led_state}"
+        )
+
+        return (
+            projector_location,
+            float(led_location.position),
+            float(led_focus.position),
+            float(laser_focus.position),
+            str(led_state),
         )
 
     async def get_laser_parameters(self) -> tuple:
--- a/python/lsst/ts/observatory/control/maintel/mtcalsys.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcalsys.py
@@ -155,9 +155,9 @@ class MTCalsys(BaseCalsys):
         )
 
         self.mtcamera = mtcamera
-        self.ls_select_led_location = 9.96  # mm
-        self.ls_select_laser_location = 79.96  # mm
-        self.led_rest_position = 120.0  # mm
+        self.linearstage_projector_locations = {"led": 9.96, "laser": 79.96}
+        self.led_rest_position = 100.0  # mm
+        self.linearstage_projector_pos_tolerance = 0.2
 
         self.laser_enclosure_temp = 20.0  # C
         self.laser_warmup = 20.0  # sec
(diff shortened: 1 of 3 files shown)


python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -155,8 +155,7 @@ def __init__(
         )
 
         self.mtcamera = mtcamera
-        self.ls_select_led_location = 9.96  # mm
-        self.ls_select_laser_location = 79.96  # mm
+        self.ls_select_locations = {"led": 9.96, "laser": 79.96}
```
tribeiro, line 158: Since you are renaming this attribute, consider expanding `ls` -> `liner_stage`.

Also, what does "select_locations" means?  Consider more expressive names that clearly indicate what this attribute is.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +340,29 @@ async def setup_laser(
         await self.change_laser_optical_configuration(optical_configuration)
         await self.change_laser_wavelength(wavelength, use_projector)
 
+    async def get_projector_setup(self) -> tuple:
```
tribeiro, line 343: When you do a type hint for a tuple with a known number of values and types, you should add that to the type annotation, e.g.:

```
async def get_projector_setup(self) -> tuple[str, float, float, float]:
```

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +340,29 @@ async def setup_laser(
...
+        """Get configuration of flatfield projector
+
+        Returns
+        -------
+            list : configuration details
```
tribeiro, line 348: The type hint says it is a tuple, but here you say it is a list. 

You are returning a tuple with known size and type so, please, update the docstring.

Also, you might want to state what each position of the tuple contains.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +340,29 @@ async def setup_laser(
...
+        """
+
+        select_location = await self.linearstage_projector_select.position
```
tribeiro, line 351: this is awaiting on `position` from `self.linearstage_projector_select`. Is this really an await-able?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -339,6 +340,29 @@ async def setup_laser(
...
+        led_location = await self.linearstage_led_select.position
+        led_focus = await self.linearstage_led_focus.position
+        laser_focus = await self.linearstage_laser_focus.position
+        led_state = await self.rem.ledprojector.evt_ledstate.aget()
```
tribeiro, line 355: You might want to add a timeout here or this might lock forever.

```
led_state = await self.rem.ledprojector.evt_ledstate.aget(timeout=self.fast_timeout)
```

tribeiro (changes requested):
Here some suggested updates. Also, you might want to check your implementation of `get_projector_setup`, I am not sure that is correct, it seems to be awaiting an attribute from remotes, but I am not sure.

Also, you might want to consider adding some unit tests..

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: No, it was written correctly to get the telemetry. Going to use tel_position.next(flush=True) as you recommended before

tests/maintel/test_mtcalsys.py
```diff
@@ -164,6 +164,10 @@ async def test_prepare_for_whitelight_flat(self) -> None:
         finally:
             self.mtcalsys.mtcamera = None
 
+        output = await self.mtcalsys.get_projector_setup()
```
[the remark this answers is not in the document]
parfa30: No problem. However, now that there is no return, I'm not completely sure what to check for besides running the script. If you have a recommendation I'd appreciate it.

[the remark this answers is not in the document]
edennihy: You could check that the method runs to completion and check that the different tel_position calls were await once.  But to be clear, I don't think you needed to remove the output. Actually, since this is get_projector_status I think it should have some output. Just thought it would be nice to log the info here instead of in the scripts. There might be another use case for the output besides the log.info.

edennihy (commented):
Thanks for all the changes, if you could add the output back and update the unit test I think it will be ready to merge.

tests/maintel/test_mtcalsys.py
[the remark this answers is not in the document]
parfa30: Ok great, it's back in!

edennihy (approved):
Looks good! Please squash your commits but otherwise ready to merge.

PR #192, opened 2025-02-28
DM-49257: added white light source tests to mtcalsys.yaml

parfa30:
It was easier to just add new sequences to this file than to change `use_camera` in a script,


tribeiro (approved):
Looks good!

PR #193, opened 2025-03-06
DM-49346: Add a park projector function

parfa30:
Also added homing into setup_calsys and LEDs OFF in prepare_for_flat. As I continue to write SALScripts, I'm finding several things that were missing/incorrect.

### DIFF
--- a/doc/news/DM-49346.feature.rst
+++ b/doc/news/DM-49346.feature.rst
@@ -1,0 +1,1 @@
+Added a function to park the LED Projector.
--- a/python/lsst/ts/observatory/control/maintel/mtcalsys.py
+++ b/python/lsst/ts/observatory/control/maintel/mtcalsys.py
@@ -130,12 +130,12 @@ class MTCalsys(BaseCalsys):
     ) -> None:
 
         self.electrometer_projector_index = 103
-        self.fiberspectrograph_blue_index = 1
-        self.fiberspectrograph_red_index = 2
-        self.linearstage_led_select_index = 1
-        self.linearstage_led_focus_index = 2
-        self.linearstage_laser_focus_index = 3
-        self.linearstage_select_index = 4
+        self.fiberspectrograph_blue_index = 101
+        self.fiberspectrograph_red_index = 102
+        self.linearstage_led_select_index = 102
+        self.linearstage_led_focus_index = 104
+        self.linearstage_laser_focus_index = 101
+        self.linearstage_select_index = 103
 
         super().__init__(
             components=[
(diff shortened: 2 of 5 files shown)


doc/news/DM-49346.feature.rst
```diff
@@ -0,0 +1 @@
+Added a function to park the LED Projector. Also added homing to the setup_calsys and turned off LEDs in the prepare_for_flat
```
tribeiro, line 1: Looks like you are adding more than one feature. Please, consider splitting them in different news fragments file.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -130,12 +130,12 @@ def __init__(
     ) -> None:
 
         self.electrometer_projector_index = 103
-        self.fiberspectrograph_blue_index = 1
-        self.fiberspectrograph_red_index = 2
-        self.linearstage_led_select_index = 1
-        self.linearstage_led_focus_index = 2
-        self.linearstage_laser_focus_index = 3
-        self.linearstage_select_index = 4
+        self.fiberspectrograph_blue_index = 101
```
tribeiro, line 133: I don't think this change was documented in the news fragment.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -439,6 +485,19 @@ async def prepare_for_flat(self, sequence_name: str) -> None:
             )
 
         if calibration_type == CalibrationType.WhiteLight:
+            # Turn off all LEDs
+            await self.rem.ledprojector.cmd_switchAllOff.set_start(
+                timeout=self.long_timeout,
+            )
+            # Confirm that Projector selection is at the LED location
+            vertical_pos = self.linearstage_projector_select.tel_position.get()
```
tribeiro, line 493: it is a bit dangerous to use `get`.. Since this is telemetry, I would suggest you use `next` with flush=True and a timeout.

```
vertical_pos = await self.linearstage_projector_select.tel_position.next(flush=True, timeout=self.long_timeout)
```

tribeiro (commented):
Looks mostly good! I only have a small concern about the use of `get` in prepare for flat. You can get into a situation where the event loop has not had a chance to read telemetry or, if telemetry is not being published, you might get outdated information. Since you are reading a telemetry, I suggest you change it to `next` with `flush=True`.

Also, please, make sure you split the difference updates you are doing in different news fragments.

edennihy (approved):
All comments addressed. PR Approved. One minor comment to consider before merging, when you rebase to the develop branch, could you please update the commit message to be a bit more specific? Combining the descriptions in the news fragments would be enough.

edennihy:
Could you please also consider adding a simple unit test for the new function? Even some basic coverage like what we provide here https://github.com/lsst-ts/ts_observatory_control/blob/f1c594df57b23d91581abe2d58bb69d8affcb6da/tests/maintel/test_mtcalsys.py#L116 would be nice given the number of different cmds that get called in this new method.

PR #194, opened 2025-03-14
Tickets/DM-49278: Add mechanism in the LSSTCam class to handle filter changes

MarcoRocchietti:


python/lsst/ts/observatory/control/maintel/lsstcam.py
```diff
@@ -222,6 +225,14 @@ async def setup_filter(
             End set filter event data.
         """
         if filter is not None:
+            if self.mtcs is not None:
```
tribeiro, line 228: I think this is pretty close, however, I would suggest you move this to a separate private method, to isolate the operation from the method here. Something like:

```
await self._setup_tcs_for_filter_change()
```

Then you put the code here inside this new method.

Also, a couple things to keep in mind. I think we want to allow the users to disable handling the rotator using the check parameter, so you might want to do something like:

```
if self.mtcs.check.mtmount and self.mtcs.check.mtptg and self.check.mtrotator:
    await self.mtcs.stop_tracking()
else:
    self.log.warning(f"Check on mtmount ({self.mtcs.check.mtmount}), mtptg ({self.mtcs.check.mtptg}) or mtrotator ({self.check.mtrotator}) disabled, changing filter without stop tracking.")

if self.mtcs.check.mtrotator:
    await self.mtcs.move_rotator(position=0.0)
else:
    self.log.warning("Rotator being ignored, skip moving rotator to filter change position.")
```

Also, I think we might want to have the position as a class attribute. We currently know it is supposed to be zero, but rather than having this magic number here we could have a `rotator_filter_change_position` attribute.

tribeiro (commented):
Looks good but I have a few suggestions. Also, I just noticed that the LSSTCam class is missing a check to see if the filter needs changing. Can you add a check like the one we [have in ComCam](https://github.com/lsst-ts/ts_observatory_control/blob/0abde6769509bfdf3ac636e97a5c6b0cf944850a/python/lsst/ts/observatory/control/maintel/comcam.py#L207-L212) to the `setup_instrument` method?

tribeiro (approved):
Looks good! thank you!

PR #200, opened 2025-04-03
tickets/DM-49506: Add open and close MTDome shutter door methods

cvillalon:
Ticket requested just open and close scripts for the MTDome shutter door but the implementation in `ts_observatory_control` was missing. This PR implements these two methods.


tribeiro (commented):
Looks great! thank you!

tribeiro (approved):

PR #202, opened 2025-04-08
DM-49553: updated led stage values

parfa30:
Updated the values for the LED stages based on this confluence page: https://rubinobs.atlassian.net/wiki/spaces/LTS/pages/50070778/Calibration+Flat-field+Screen+Central+Section+Illumination+Hardware

### DIFF
--- a/doc/news/DM-49553.feature.1.rst
+++ b/doc/news/DM-49553.feature.1.rst
@@ -1,0 +1,1 @@
+Updated the locations for the LED stages.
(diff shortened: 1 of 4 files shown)


parfa30:
This will not pass until https://github.com/lsst-ts/ts_config_mtcalsys/pull/36 is merged.

elanaku (approved):
I think this looks good. One thing is that I think each news feature should be just one item/sentence. So you should either make it one sentence, or make two news fragments, one called DM-49553.feature.1.rst and the other called DM-49553.feature.2.rst.

parfa30:
Thanks, I split the news fragment in two and squashed commits

PR #205, opened 2025-04-16
DM-50224: added mtcamera_filter to whitelight_r_source

parfa30:


parfa30:
Now using this ticket as a catchall for a bunch of mtcalsys changes needed.

python/lsst/ts/observatory/control/data/mtcalsys.yaml
elanaku, line 1: I think the mtcamera filter may need to be the physical filter, not the band. In which case it would need to be r_57

elanaku (commented):
Looks good. I think the one thing is that we may need to list the physical filter, not just the band.

elanaku (approved):
Looks good now!

PR #206, opened 2025-04-16
Avoid change filter command if filter is in place

gmegh:


tribeiro:
This is already handled [here](https://github.com/lsst-ts/ts_observatory_control/blob/3ec436bfa23c10f0edc78ec89f750206dff451ab/python/lsst/ts/observatory/control/maintel/lsstcam.py#L248-L251) as I have told you before.

PR #209, opened 2025-04-22
DM-50396: added groupId to electrometer and fiberspec exposures

parfa30:


python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -851,19 +851,30 @@ async def _take_data(
         )
         exposures_done: asyncio.Future = asyncio.Future()
 
+        group_id = None
```
edennihy, line 854: I think group_id has to be a str value for the cmd expose in the fiberspectrograph (https://ts-xml.lsst.io/sal_interfaces/FiberSpectrograph.html#groupid) and electrometer (https://ts-xml.lsst.io/sal_interfaces/Electrometer.html#electrometer-commands-startscandt-groupid), so you should probably set this up to default to an empty string. 

You could do something like this:
`group_id = exposure_metadata.get("group_id", "")

if group_id == "":
    self.log.warning(
        "No Group ID for Electrometer/Fiber Spectrograph exposures. Continuing"
    )
`

edennihy (approved):
Approved, but please see my comment above before merging.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -951,6 +964,7 @@ async def take_fiber_spectrum(
         self,
         fiberspectrograph_color: str,
         exposure_time: float | None,
+        group_id: str | None,
```
edennihy, line 967: If you adopt the comment, please update this to be just a str.

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -900,6 +911,7 @@ async def _take_data(
     async def take_electrometer_scan(
         self,
         exposure_time: float | None,
+        group_id: str | None,
```
edennihy, line 914: If you adopt the comment, please update this to be just a str.

PR #210, opened 2025-04-25
Tickets/dm 49954 merge

edennihy:


tribeiro (changes requested):
Sorry but there are some changes that were included here that need to be undone:

- Commit 71459cd66d4cedd29181af44a9ca79ad29da6303 reverts a change that should not be reverted. Unfortunately, removing this commit also means removing some other commits further down the line to undo the unit tests changes. Those should also be removed but I didn't got their hash.
- Commit fbf8caf69e3f77d84549d864920d0af0729d2f55 should also be removed, this has a work around to the rotator not moving or publishing in position when it is close to the commanded position. It is a dirty work around and should not be included.

Some commits includes a bunch of "empty lines removal". They are probably due to a bad linter configuration reformatting the files upon commit. It would be great if you could check those out because otherwise it will be a fight between lint versions (one person add another  person removes).

tribeiro (approved):

PR #223, opened 2025-06-02
tickets/DM-50762: Move MTAOS start/stop close loop logic to MTCS

cvillalon:


python/lsst/ts/observatory/control/maintel/mtcs.py
edennihy, line 1: In order to be able to use this in a SAL script, we will need to add the startClosedLoop, stopClosedLoop and closedLoopState to one of the Usage sets. 

Since this is an AOS specific method with no other dependencies, I recommend we add it to the AOS usages here https://github.com/lsst-ts/ts_observatory_control/blob/7e0f0c76f0bfc64d76cf51730560548add85b4fe/python/lsst/ts/observatory/control/maintel/mtcs.py#L3642.

edennihy (changes requested):
One minor (but important) comment to consider. Otherwise, it looks good!

edennihy (approved):
Approved.

PR #226, opened 2025-06-06
DM-51282: Changing how PTC curves are generated

parfa30:


python/lsst/ts/observatory/control/data/mtcalsys.yaml
edennihy, line 1: I think these updates to mtcalsys.yaml should be moved other to other PR (https://github.com/lsst-ts/ts_observatory_control/pull/219)

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -905,41 +909,56 @@ async def calculate_optimized_exposure_times(
...
-                or config_data["use_cbp_electrometer"],
-            )
-            fiberspectrograph_exptimes_red = (
-                await self._calculate_fiberspectrograph_exposure_times(
-                    exptimes=config_data["exposure_times"],
-                    use_fiberspectrograph=config_data["use_fiberspectrograph_red"],
+            for exptime in config_data["exposure_times"]:
+                dac = config_data["dac_value"]
+                if config_data["ptc"]:
+                    if exptime < 1.0:
```
edennihy, line 921: This could lead to some weird behavior...for example if exptime=0.99s, it would reset exptime to 99s. Is that okay?

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -905,41 +909,56 @@ async def calculate_optimized_exposure_times(
...
+                        dac = 0.15
+                        exptime = exptime * 100.0
+                    elif exptime > 30.0:
```
edennihy, line 924: This could also result in an exptime that is negative or less than 1s. As written, if exptime=31s, then it would change to -9 sec. Also if it is 40.1s, it would become 0.1s. I don't think either of those are what we want.

edennihy (changes requested):

python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -906,12 +911,66 @@ async def run_calibration_sequence(
...
+            samples_per_bin.
+        dac_values : `list`
+        Raises
+        ------
+        ValueError
+            If fewer than two bin edges are provided.
+        """
+        exposure_times_list: list[NDArray[np.float_]] = []
+        bin_edges = np.array(bin_edges, dtype=float)
+
```
edennihy, line 953: You could consider adding a check in there that dac_values is len(bin_edges) - 1. You can do that with an assert statement that raises an info message for the user when they are not.

parfa30: It should actually be the length of bin_edges, because bin_edges is now a list of 2 items: [start, end]. But I will add and assert.

PR #227, opened 2025-06-06
Tickets/dm 51020: Add function to create a daily PTC sequence in mtcalsys with random exp_times.

edennihy:


python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -186,6 +186,88 @@ def __init__(
             fiberspectrograph=0.0,
         )
 
+    def get_calibration_configuration(self, name: str) -> dict[str, typing.Any]:
```
parfa30, line 189: Could we just incorporate this into `calculate_optimized_exposure_times`

python/lsst/ts/observatory/control/data/mtcalsys_schema.yaml
```diff
@@ -159,6 +159,27 @@ properties:
     type: array
     items:
       type: number
+  constrained_random_exposure_times:
```
parfa30, line 162: Could we add `dacValue` to this configuration? Most of the bins will have different LED brightness, which correlated with dacValue

python/lsst/ts/observatory/control/maintel/mtcalsys.py
parfa30: If you want to keep this separate, could we make the name of the function a bit more specific?

parfa30 (changes requested):

PR #228, opened 2025-06-06
OSW-494: Update m1m3/check_actuator.py script to run bump tests in parallel.

tribeiro:


dsanmartim (approved):
Approved. I just added a small correction needed in the news fragment. And I also have a question about the minimal distance:

Do you see any edge case (e.g., event delivery delays, incorrect actuator state reporting, or concurrency timing issues) that could lead to a violation of the bump_test_minimal_distance rule when running tests in real operation?

PR #235, opened 2025-07-16
Tickets/dm 51639: Support Summit Observing Weeks 27-28 of 2025 

edennihy:


dsanmartim (approved):
This looks good to me... approved! 

However, I was wondering — would it make sense to squash all the commits related to a specific news fragment (and the associated file changes), especially the more generic ones like “fix bugs in CBP method”? Is there a particular reason to keep the full history of these incremental changes?

PR #245, opened 2025-09-08
BLOCK-T339: added a setup_cbp to mtcalsys.yaml

parfa30:


python/lsst/ts/observatory/control/data/mtcalsys.yaml
```diff
@@ -1,6 +1,25 @@
 # Configuration for calibration hardware setup
 # Created 2024-07-15
 
+setup_cbp:
```
elanaku, line 4: set use_camera to false.

elanaku (approved):
Looks good!

PR #293, opened 2026-03-11
Tickets/osw 1968

edennihy:


python/lsst/ts/observatory/control/maintel/mtcs.py
```diff
@@ -1024,13 +1024,22 @@ def set_azel_slew_checks(self, wait_dome: bool) -> typing.Any:
...
+        *,
+        timeout: float | None = None,
+    ) -> None:
         """Utility method to slew dome to a specified position.
 
         Parameters
         ----------
         az : `float` or `str`
             Azimuth angle for the dome (in deg).
+        timeout : `float`
```
dsanmartim: Similar to the docstring in base_tcs.py, here it should be ``timeout : `float`, optional`` (you missed the optional).

python/lsst/ts/observatory/control/auxtel/atcs.py
```diff
@@ -329,6 +335,9 @@ async def slew_dome_to(self, az: float, check: typing.Any = None) -> None:
         ----------
         az : `float` or `str`
             Azimuth angle for the dome (in deg).
+        timeout : `float`
```
dsanmartim: Similar to the docstring in base_tcs.py, here it should be ``timeout : `float`, optional`` (you missed the optional).

dsanmartim: Just a minor thing. The order of the parameters in the docstring is inverted when compared to the order they appear in the function definition (`check` appears before `timeout`). This also happens with the `mtcs.py`.

dsanmartim (approved):
Approved. Just pointed out some really cosmetic "improvements".

PR #295, opened 2026-03-16
DM-51498: Electrometer and FiberSpec now optional

parfa30:
In order to toggle the use of the electrometer and fiber spectrographs from external scripts, had to make some changes to MTCalsys.

- removed `use_fiberspectrograph_blue` and `use_fiberspectrograph_red` from the configuration file (`mtcalsys.yaml`)
- Added `use_electrometer` and `use_fiberspectrograph_blue` and `use_fiberspectrograph_red` to the intiation of MTCalsys, which can be overwritten in an external script
- used configuration file for identifying which electrometers to use (projector/flatfield, cbp, cbp_cal). These are defined in `setup_calsys` and also `take_electrometer_scan`


edennihy (approved):
Approved.

PR #299, opened 2026-04-19
swapped F1 and F2 for CPB and Mono

parfa30:
When the Stubbs Laser was installed on the dome, the fibers to the CBP and Flatfield Projector were swapped this is now updated in mtcalsys.yaml


kakoon:
Line 252 has a typo, other than that, looks good!

kakoon (approved):
Approved!

PR #300, opened 2026-04-19
adding cbp_u_4nm for diagnostic testing

parfa30:
Reducing number of wavelengths so that the test lasts ~10minutes


kakoon (approved):
looks good!

PR #311, opened 2026-05-27
updates to get_laser_parameters in mtcalsys

parfa30:


python/lsst/ts/observatory/control/maintel/mtcalsys.py
```diff
@@ -624,23 +624,53 @@ async def get_projector_setup(
...
-        return await asyncio.gather(
-            self.rem.tunablelaser.evt_opticalConfiguration.aget(
-                timeout=self.long_timeout
-            ),
-            self.rem.tunablelaser.evt_wavelengthChanged.aget(timeout=self.long_timeout),
-            self.rem.tunablelaser.evt_interlockState.aget(timeout=self.long_timeout),
-            self.rem.tunablelaser.evt_burstModeSet.aget(timeout=self.long_timeout),
-            self.rem.tunablelaser.evt_continuousModeSet.aget(timeout=self.long_timeout),
+    async def get_laser_parameters(self) -> tuple:
+        """Get laser configuration."""
```
edennihy: Can you add a doctring here for the Tuple that this returns? It should say that the type could be Any or None if the .aget fails to return. Could also mention that this is needed because the TunableLaser CSC does not always publish these events.

edennihy (approved):
Approved, but please see my comment about adding an updated docstring to the method.

PR #312, opened 2026-05-30
Updated CalSys scans and bug fix

parfa30:


b1quint (approved):
I see that you also improved some of the logic and docs. Cool! Thanks! It looks good!

PR #321, opened 2026-06-10
Tickets/osw 1626: Add ATBuilding class

edennihy:


edennihy:
Tested on bts on 2026-06-11.

doc/news/OSW-1626.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``ATBuilding`` class with vent gate control methods and extraction fan control methods and ``ATBuildingAsyncMock`` for unit testing.
```
tribeiro: Just a reminder that news fragments should be made in the past tense, e.g. "Added ..."

python/lsst/ts/observatory/control/auxtel/atbuilding.py
```diff
@@ -0,0 +1,377 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
```
tribeiro, line 21: This header is missing the last line:

```
# along with this program. If not, see <https://www.gnu.org/licenses/>.
```

python/lsst/ts/observatory/control/auxtel/atbuilding.py
```diff
@@ -0,0 +1,377 @@
...
+            If waiting for the event times out.
+        """
+        if timeout is None:
+            timeout = self.long_timeout
+
+        pending = set(gates)
+
+        try:
+            while pending:
+                vent_gate_state = await asyncio.wait_for(
```
tribeiro: Instead of using `asyncio.wait_for` you can use the `timeout` parameter in the `next` method:

```
await self.rem.atbuilding.evt_ventGateState.next(flush=False, timeout=timeout)
```

also, you need to be careful when using `flush=False` in this context. The thing is, if you had not cleared the queue before you sent the command that you expect will change this event, you might get outdated information.

The safest way to handle is to do a `flush` first, then an `aget` then iterate using `next`. Something like this:

```
self.rem.atbuilding.evt_ventGateState.flush()

vent_gate_state = await self.rem.atbuilding.evt_ventGateState.aget(timeout=timeout)

pending = {gate for gate in vent_gate_state.state if VentGateState(vent_gate_state.state[gate]) != expected_state}

while pending:
    try:
        vent_gate_state = await self.rem.atbuilding.evt_ventGateState.next(flush=False, timeout=timeout)
    except asyncio.TimeoutError:
        raise RuntimeError(
            f"Timed out waiting for vent gates {sorted(pending)} "
            f"to reach state {expected_state.name!r}."
        )        
    pending = {gate for gate in vent_gate_state.state if VentGateState(vent_gate_state.state[gate]) != expected_state}

```

A few things to note;

1. I compacted the loop to retrieve the "pending" gates.
2. I always check all the gates, which allow us to capture cases when a gate go in and out of the expected state. You might want to review this part but it might be useful.

python/lsst/ts/observatory/control/auxtel/atbuilding.py
```diff
@@ -0,0 +1,377 @@
...
+    def usages(self) -> typing.Dict[int, UsagesResources]:
+        if self._usages is None:
+            usages = super().usages
+
+            usages[self.valid_use_cases.Setup] = UsagesResources(
+                components_attr=["atbuilding"],
+                readonly=False,
+                generics=["summaryState"],
+                atbuilding=[
+                    "openVentGate",
```
tribeiro: You don't need to define the command here. I know some old usage definitions do this but it was never necessary. Only events and telemetry need to be specified.

python/lsst/ts/observatory/control/auxtel/atbuilding.py
```diff
@@ -0,0 +1,377 @@
...
+                    "closeVentGate",
+                    "setExtractionFanDriveFreq",
+                    "ventGateState",
+                ],
+            )
+
+            usages[self.valid_use_cases.DryTest] = UsagesResources(
+                components_attr=(), readonly=True
+            )
+
```
tribeiro, line 142: you might want to add a StandardStateTransition usage and include heartbeats and summaryState events.

tribeiro (approved):
Looks good! I left a few comments for you to consider...

PR #328, opened 2026-06-26
ability to modify configurations in MTCalsys

parfa30:


parfa30:
To work with https://github.com/lsst-ts/ts_externalscripts/pull/239. Tested on BTS. This will allow us to make changes to tests without having to push changes to the summit. It will also enable us to delete many sequences in mtcalsys.yaml that are very similar but not identical.

elanaku (approved):
Looks good.

